PR #1, opened 2019-03-16
Initial experiment conventions

piercefreeman:


CHANGES.md
```diff
@@ -0,0 +1,3 @@
+# Release Change Log
+
+Update this file when creating new releases, with most recent releases first.
```
adamhadani: we dont typically update CHANGELOG files from what I've seen elsewhere in microcosm land. might be annoying to just have this file in there if not updated. I'm for removing unless we're diligent about updating with major releases

piercefreeman: I've seen it in a few places, but you're right - doesn't look commonly updated.  I'll remove.

microcosm_sagemaker/loaders.py
```diff
@@ -0,0 +1,93 @@
...
+    if base_configuration_url:
+        remote_configuration = load_from_s3(base_configuration_url)(metadata)
+
+        # Locally specified hyperparameters should take precedence over the
+        # base configuration
+        new_configuration = Configuration()
+        new_configuration.merge(remote_configuration)
+        new_configuration.merge(configuration)
+
+        configuration = new_configuration
```
pokey: I think you could replace all these lines with 

```py
configuration = merge(
    remote_configuration,
    configuration,
)
```

microcosm_sagemaker/mixins/base.py
```diff
@@ -0,0 +1,71 @@
...
+        Predict the value of some datapoint, specified by the *args and **kwargs.
+        """
+        pass
+
+    def _set_constant_seed(self, constant=STATIC_SEED):
+        seed(constant)
+
+        try:
+            from torch import manual_seed
+            manual_seed(constant)
```
pokey: Not in love with putting `torch` stuff in this base class.  I think that a derived `TorchBundle` should handle that

microcosm_sagemaker/mixins/base.py
```diff
@@ -0,0 +1,71 @@
...
+        except ModuleNotFoundError:
+            pass
+
+        try:
+            from tf.random import set_random_seed
```
pokey: As above

microcosm_sagemaker/mixins/base.py
```diff
@@ -0,0 +1,71 @@
...
+            set_random_seed(constant)
+        except ModuleNotFoundError:
+            pass
+
+    def _save_environment(self, artifact_dir):
+        configuration_path = Path(artifact_dir) / "configuration.json"
+        with open(configuration_path, "w") as configuration_file:
+            dump(self._environment, configuration_file)
```
pokey: Similar to my comment above re `json.load`, I'd argue for

```suggestion
            json.dump(self._environment, configuration_file)
```

microcosm_sagemaker/mixins/base.py
```diff
@@ -0,0 +1,71 @@
...
+    @abstractmethod
+    def fit(self, artifact_path):
+        """
+        Train a model.
+
+        :param artifact_path: {str} location to place model artifacts
+        :param configuration: {dict} of configuration values
+        """
+        self._set_constant_seed()
+        self._save_environment(artifact_path)
```
pokey: The problem with putting logic like this into an abstract method is that people might forget to call the base method.  I see two immediate alternatives:
- Define a separate abstract `_fit` for derived classes to override and call it here, or
- Do the seeding / setup in a separate method that is called by the client of this class.

setup.py
```diff
@@ -0,0 +1,44 @@
...
+    setup_requires=[
+        "nose>=1.3.6",
+    ],
+    dependency_links=[
+    ],
+    entry_points={
+        "console_scripts": [
+            "train = microcosm_sagemaker.commands.train:train_cli",
+            "evaluate = microcosm_sagemaker.commands.evaluate:evaluate_cli",
+            "runserver = microcosm_sagemaker.commands.runserver:runserver_cli",
```
pokey: In `microcosm-flask` and `microcosm-postgres`, we define the functions but leave it to the importing code to define the console scripts.  I am not fully privy to the reasoning there but I imagine it could be a bit annoying if your libraries are defining script entrypoints underneath you.  OTOH, if we plan to use these standard commands in CI maybe we want to enforce this.  idk cc/ @adamhadani

setup.py
```diff
@@ -0,0 +1,44 @@
...
+        ],
+        "microcosm.factories": [
+            "active_bundle = microcosm_sagemaker.factories:configure_active_bundle",
```
pokey: Where is this used?

microcosm_sagemaker/commands/evaluate.py
```diff
@@ -0,0 +1,38 @@
...
+from microcosm_sagemaker.app_hooks import AppHooks
+
+
+def evaluate(input_path, artifact_path):
+    graph = AppHooks.create_serve_graph(artifact_path=artifact_path)
+
+    chdir(input_path)
+
+    # Load the saved artifact
+    graph.bundle_manager.load(artifact_path)
```
pokey: Where is `bundle_manager` defined?

microcosm_sagemaker/loaders.py
```diff
@@ -0,0 +1,93 @@
...
+    When we train a model, we freeze all of the current graph variables and store it alongside
+    the artifact. Whenever we boot up the model again, we want to hydrate this from disk.
+
+    """
+    def _load_path(metadata):
+        path = join(artifact_path, "configuration.json")
+
+        try:
+            with open(path) as raw_file:
+                return load(raw_file)
```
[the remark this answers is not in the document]
piercefreeman: Fair point. In the interest of consistency, I'll keep the function import but I'll rename it to make clear the json origin.

Something like: `from json import load as json_load, loads as json_loads`

setup.py
piercefreeman: There was some legacy code that mixed `active_bundle` with `bundle_manager`.  The correct one should be `active_bundle`, which is defined in `/factories.py`.  It picks up on the user specified model that should be active during training and inference.

microcosm_sagemaker/loaders.py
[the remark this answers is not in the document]
pokey: That's probably fine, but curious to know what you mean by "In the interest of consistency".  I don't think we have any policy against `import foo`, do we?  cc/ @adamhadani

README.md
```diff
@@ -9,67 +9,67 @@ After creating a new model, there are a few steps to integrate with microcosm-sa
...
+from microcosm_sagemaker.loaders import load_train_conventions
 
-    def create_app(debug=False, testing=False, extra_config={}):
-      config_loader = load_each(
-          load_from_dict(extra_config),
-          load_from_environ,
-          load_train_conventions,
-      )
+  def create_app(debug=False, testing=False, extra_config={}):
+    config_loader = load_each(
```
pokey: I'd still argue for 4 spaces indent here

microcosm_sagemaker/loaders.py
```diff
@@ -0,0 +1,93 @@
...
+    base_configuration_url = configuration.pop("base_configuration", None)
+
+    if base_configuration_url:
+        remote_configuration = load_from_s3(base_configuration_url)(metadata)
+
+        # Locally specified hyperparameters should take precedence over the
+        # base configuration
+        configuration = merge([
+            remote_configuration,
+            configuration
```
pokey:
```suggestion
            configuration,
```

README.md
```diff
@@ -1,2 +1,75 @@
...
+    return graph.lock()
+    ```
+
+3. Expose the graphs in `setup.py`.
+
+    ```
+  setup(
+    name="my_model",
+    entry_points={
+      "microcosm_sagemaker.graph_hooks": [
```
pokey: We have a quasi-convention to name the entry point the same as the module where it's used, which would argue for `microcosm_sagemaker.app_hooks`

README.md
piercefreeman: Maybe I'm misinterpreting, but it looks like this line already has a 4 space indent?

microcosm_sagemaker/commands/runserver.py
```diff
@@ -0,0 +1,23 @@
...
+
+@command()
+@option(
+    "--artifact_path",
+    type=Path(),
+    required=True,
+    help="Path for reading artifacts, used for local testing",
+)
+def runserver_cli(artifact_path):
+    graph = AppHooks.create_serve_graph(debug=True, artifact_path=artifact_path)
```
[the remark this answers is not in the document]
piercefreeman: @pokey For this particular case, we need to provide the artifact path to the graph before it's created so we can re-hydrate its configuration back into the graph.  We could force every client to do that CLI parsing manually (before it does the graph creation), but it feels like something that's a better fit for the core conventions of the library. WDYT?

setup.py
piercefreeman: Let's continue this conversation in the comment above on `runserver_cli`.

README.md
pokey: Haha I can't argue with that statement.  What I meant was that we want to use 4-space indents, so the prev line would have 4 spaces, and this line would have 8

microcosm_sagemaker/commands/train.py
```diff
@@ -44,6 +45,9 @@ def train_cli(configuration, input_path, artifact_path, auto_evaluate):
     if not input_path:
         input_path = SagemakerPath.INPUT
 
+    artifact_path = abspath(artifact_path)
```
pokey: I'd argue for using the `resolve_path` argument to [`click.Path`](http://click.palletsprojects.com/en/7.x/api/#click.Path), and any other `Path` arguments that make sense, such as `dir_okay`, `exists`, etc.  Also, not sure it's possible in this case, but I always try to use `File` when I can to let `click` handle opening the file.

microcosm_sagemaker/metrics/models.py
```diff
@@ -0,0 +1,36 @@
...
+@unique
+class MetricMode(Enum):
+    TRAINING = "Training"
+    DEVELOP = "Develop"
+
+
+@unique
+class MetricUnit(Enum):
+    """
+    In-exhaustive list of common CloudWatch metric units
```
pokey:
```suggestion
    Inexhaustive list of common CloudWatch metric units
```

microcosm_sagemaker/metrics/store.py
```diff
@@ -0,0 +1,48 @@
...
+
+from boto3 import client
+from botocore.exceptions import ClientError
+from microcosm.api import binding
+
+from microcosm_sagemaker.metrics.models import MetricMode
+
+
+@binding("sagemaker_metrics")
+class SMMetrics(object):
```
pokey:
```suggestion
class SageMakerMetrics(object):
```

PR #17, opened 2019-04-15
GLOB-30893 Refactor

pokey:


microcosm_sagemaker/conventions/invocations.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    graph.config.swagger_convention.operations.append("saved_search")
+    ns = Namespace(
+        subject="invocations",
+        version=None,
+    )
+    mappings = {
+        Operation.SavedSearch: definition,
+    }
+    configure_saved_search(graph, ns, mappings)
```
piercefreeman: Are we saving this this search anywhere?  Seems like a slight misnomer in this context.

microcosm_sagemaker/testing/directory_comparison.py
```diff
@@ -0,0 +1,50 @@
...
+from hamcrest import (
+    assert_that,
+    contains,
+    equal_to,
+    is_,
+)
+from hamcrest.core.base_matcher import BaseMatcher
+
+
+def directory_comparison(gold_dir: Path,
```
piercefreeman: I normally newline indent function definitions that are too long:

```
def directory_comparison(
    gold_dir: Path,
    actual_dir: Path,
    matchers: Optional[Mapping[Path, BaseMatcher]] = None
)
```

microcosm_sagemaker/testing/json_matcher.py
```diff
@@ -0,0 +1,24 @@
...
+        return self.matcher.matches(json.loads(raw_bytes.decode()))
+
+    def describe_to(self, description):
+        description.append_text('JSON matches ')
+        self.matcher.describe_to(description)
+
+
+def json_matches(matcher):
+    """
+    When matching, it expects raw bytes, which it will decode, json.loads, and
```
piercefreeman: When we typically handle json, we rely on the test unit writer to convert it to a dictionary and then compare its values. In what situations would this be preferable to that more explicit approach?

microcosm_sagemaker/conventions/invocations.py
pokey: Yes I had the same reaction when I saw this terminology.  I was merely repeating what was already extant in `ml-taxonomysimilarity-classifier`, but I believe what's happening is that `SavedSearch` is the convention that we use when we have a search that expects a `POST` instead of a `GET`.  A bit confusing.  Maybe I'll just add a NB

microcosm_sagemaker/testing/directory_comparison.py
pokey: So my understanding is that we follow the convention that I've used here, but I don't think that convention is formalized anywhere.  Just wrote up a [discussion page](https://globality.atlassian.net/wiki/spaces/DEV/pages/868057693/Newlines+in+function+definition+parameter+lists) on Confluence.

microcosm_sagemaker/testing/json_matcher.py
pokey: Let me revisit this.  I had an idea for a better approach here actually but then forgot so thanks for the reminder 😄

PR #33, opened 2019-05-07
Add artifact subdir

pokey:
This PR automatically creates a nested directory in the output artifact for the bundle to use.  This way the bundle won't interfere with any files we put at the top level of the output artifact, such as the `configuration.json`

### DIFF
--- a/microcosm_sagemaker/commands/evaluate.py
+++ b/microcosm_sagemaker/commands/evaluate.py
@@ -6,7 +6,7 @@ from click import command
 from microcosm.object_graph import ObjectGraph
 
 from microcosm_sagemaker.app_hooks import create_evaluate_app
-from microcosm_sagemaker.artifact import InputArtifact
+from microcosm_sagemaker.artifact import RootInputArtifact
 from microcosm_sagemaker.commands.options import input_artifact_option, input_data_option
 from microcosm_sagemaker.input_data import InputData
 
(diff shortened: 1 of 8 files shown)


microcosm_sagemaker/commands/evaluate.py
```diff
@@ -25,10 +25,11 @@ def main(input_data, input_artifact):
 def run_evaluate(
     graph: ObjectGraph,
     input_data: InputData,
-    input_artifact: InputArtifact,
+    input_artifact: RootInputArtifact,
 ) -> None:
     # Load the saved artifact
-    graph.active_bundle.load(input_artifact)
+    active_bundle_input_artifact = input_artifact / graph.config.active_bundle
```
salmanmashayekh: You are assuming the subdir name is the bundle name in here, but we can arbitrarily name the subdir in the cookiecutter (e.g., `get_fixture_path("example_input_artifact") / "example_bundle"`). Can we enforce the cookiecutter to use the bundle name also?

microcosm_sagemaker/commands/runserver.py
```diff
@@ -44,7 +44,8 @@ def run_serve(
         host: str,
         port: int,
 ) -> None:
-    graph.active_bundle.load(input_artifact)
+    active_bundle_input_artifact = input_artifact / graph.config.active_bundle
+    graph.active_bundle.load(active_bundle_input_artifact)
```
salmanmashayekh: This could become strange. In the `ml-nlp-serviceoffering-classifier`, the `active_bundle` is a manager bundle that sits above the two main bundles and does not have any artifacts of its own. Calling `load` on that bundle with the `active_bundle_input_artifact` requires the manager bundle to call the two main bundles with their own subdir artifacts, which are parallel to the `active_bundle_input_artifact`:
```
main_dir
  |- manager_bundle_subdir
  |- main_bundle_1_subdir
  |- main_bundle_2_subdir
```

Why don't we call the `load` on each bundle with the `root` path and each bundle can find its own subdir?

microcosm_sagemaker/commands/evaluate.py
pokey: This shouldn't actually be a problem, as described in new cookiecutter PR

pokey:
Folded into #37

PR #35, opened 2019-05-13
Add pickle comparison to the library

salmanmashayekh:


microcosm_sagemaker/testing/pickle_comparison.py
```diff
@@ -0,0 +1,95 @@
...
+        if isinstance(obj, dict):
+            return obj
+        return vars(obj)
+
+
+def pickle_extractor(binary):
+    return joblib.load(BytesIO(binary))
+
+
+def _matches_with_object(obj, ignore_properties=None):
```
pokey:
```suggestion
def matches_object(
    obj: Union[object, dict],
    ignore_properties: Optional[List[str]]=None,
):
```

I think this could be generally useful

microcosm_sagemaker/testing/pickle_comparison.py
```diff
@@ -0,0 +1,95 @@
...
+from hamcrest.core.base_matcher import BaseMatcher
+from sklearn.externals import joblib
+
+from microcosm_sagemaker.testing.bytes_extractor import ExtractorMatcherPair
+
+
+class IsObjectEqual(BaseMatcher):
+    """
+    Matcher to compare the object blob saved at filepath A with another object blob saved
+    at filepath B.
```
pokey: I don't think this line is accurate.  The objects are passed in directly, not the paths

microcosm_sagemaker/testing/pickle_comparison.py
```diff
@@ -0,0 +1,95 @@
...
+    Sometimes there is some randomness within these generated files, so the user can optionally
+    provide a `ignore_properties` array which will ignore mismatches of those given object
+    properties.  For instance, if we have:
+    MyObject(foo=bar, extras=Test)
+    MyObject(foo=bar, extras=Test2)
+    If ignore_properties is not specified, these will be unequal.  If ignore_properties is
+    set to ["extras"], they will be equal.
+    """
+    def __init__(self, obj, ignore_properties=None):
+        self.ignore_properties = ignore_properties
```
pokey: Need to test for `None` and construct an empty list

pokey: Also would be nice to add type annotations

microcosm_sagemaker/testing/pickle_comparison.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def _matches_with_object(obj, ignore_properties=None):
+    if ignore_properties is None:
+        ignore_properties = []
+    return IsObjectEqual(obj, ignore_properties)
+
+
+def matches_with_object(obj):
+    return partial(_matches_with_object, ignore_properties=[])
```
pokey: This function doesn't look right.  Have you tried just passing in `_matches_with_object` itself?

PR #37, opened 2019-05-15
GLOB-32441 Add bundle dependencies; add tests

pokey:
See [GLOB-32441](https://globality.atlassian.net/browse/GLOB-32441)


microcosm_sagemaker/dependency_traverser.py
```diff
@@ -0,0 +1,74 @@
...
+
+from microcosm_sagemaker.exceptions import DependencyCycleError
+
+
+T = TypeVar('T')
+
+
+def traverse_component_and_dependencies(
+    component: T,
+    dependency_getter: Callable[[T], Iterable[T]] = attrgetter("dependencies"),
```
piercefreeman: Will there ever be a case where we _don't_ want to use "dependencies" as the attribute to access?  Might make more sense to conventionalize within the function itself versus as a function parameter.

microcosm_sagemaker/dependency_traverser.py
```diff
@@ -0,0 +1,74 @@
...
+
+    yield component
+    seen.add(component)
+
+
+@contextmanager
+def _reserve(reserved: Set[T], component: T):
+    """
+    Adds component to a list of reserved components, erroring if the component
+    has already been reserved, then remove compenet from list of reserved
```
piercefreeman: s/compenet/component

microcosm_sagemaker/dependency_traverser.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+
+def _traverse_helper(
+    dependency_getter: Callable[[T], Iterable[T]],
+    seen: Set[T],
+    reserved: Set[T],
+    component: T,
+):
+    if component in seen:
```
piercefreeman: In which cases will `seen` and `reserved` differ?  Can we add a docstring about the edge cases?

microcosm_sagemaker/graph.py
```diff
@@ -0,0 +1,6 @@
+def get_component_name(graph, component):
```
piercefreeman: Since this is only used in one other file, I don't think we need to refactor it quite yet.  Let's put back into `bundle_traversal.py`.

microcosm_sagemaker/dependency_traverser.py
```diff
@@ -1,52 +1,53 @@
...
     Given a component in a dependency graph, traverses the graph in topological
     order, ie such that all dependencies of a component will be yielded before
     the component itself.
 
-    """
-    seen: Set[T] = set()
-    reserved: Set[T] = set()
+    Each component is expected to have a `dependencies` property that returns
+    an iterable of dependency components.  Each component is also expected to
+    be hashable.
```
pokey: I tried to formalize this using protocols, but looks like that stuff isn't mature enough yet.

microcosm_sagemaker/artifact.py
```diff
@@ -8,40 +9,48 @@
...
 class RootInputArtifact(InputArtifact):
     def __init__(self, path: Union[str, Path]):
         super().__init__(path)
 
+    def __truediv__(self, name: str):
+        return BundleInputArtifact(self.path / name)
+
     def load_config(self, metadata: Metadata) -> Configuration:
         """
         When we train a model, we freeze all of the current graph variables and store it alongside
```
pokey: I made the artifact contract stricter / more explicit.  Now:
- you can only get nested artifacts of the root artifact, ie can only nest one level deep
- we have explicit types to represent bundle artifacts

I've also updated variable names / keyword args in many places to always use `bundle_xxx_artifact` or `root_xxx_artifact` to indicate what kind of artifact it is, rather than just `xxx_artifact`

microcosm_sagemaker/bundle_orchestrator.py
```diff
@@ -14,29 +14,37 @@ def __init__(self, graph: ObjectGraph):
     @abstractmethod
     def __call__(
         self,
-        active_bundle: Bundle,
+        bundle: Bundle,
         bundle_handler: Callable[[Bundle], None],
+        dependencies_only: bool = False,
```
pokey: Needed `dependencies_only` for testing

microcosm_sagemaker/bundle_traversal.py
```diff
@@ -1,53 +1,79 @@
...
+from microcosm_sagemaker.artifact import RootInputArtifact, RootOutputArtifact
+from microcosm_sagemaker.bundle import Bundle
 from microcosm_sagemaker.input_data import InputData
 
 
 @defaults(
     bundle_orchestrator="single_threaded_bundle_orchestrator",
 )
-class ActiveBundleAndDependenciesLoader:
+class BundleAndDependenciesLoader:
```
pokey: These now support arbitrary bundles rather than just the `active_bundle`.  Needed this for testing

microcosm_sagemaker/commands/evaluate.py
```diff
@@ -16,19 +16,26 @@
 @input_artifact_option()
 def main(input_data, input_artifact):
     graph = create_evaluate_app(
-        loaders=[input_artifact.load_config],
+        extra_loader=input_artifact.load_config,
```
pokey: I've standardized on `extra_loader` for all `create_app` signatures instead of `extra_config` or `loaders`. Both of these functionalities can be recovered using `load_from_dict` and `load_each`, respectively. Within `microcosm`, it is standard to just have a single loader param and let the client use `load_each` as necessary.

microcosm_sagemaker/testing/bundle.py
```diff
@@ -0,0 +1,98 @@
+import tempfile
```
pokey: Took a swing at generic bundle test case.  Pretty happy with how it turned out, but will see if it's too restrictive for real test cases.  Clients aren't required to use it

microcosm_sagemaker/tests/app_hooks/evaluate/app.py
```diff
@@ -16,23 +16,21 @@
 
 
 Loader = Callable[[Metadata], Configuration]
+empty_loader = load_each()
```
pokey: Both of these lines would be a candidate for adding to `microcosm`, but for now they're simple and somewhat illustrative

microcosm_sagemaker/tests/app_hooks/train/app.py
```diff
@@ -2,37 +2,38 @@
...
 
+Loader = Callable[[Metadata], Configuration]
+empty_loader = load_each()
+
+
 def create_app(
     debug: bool = False,
     testing: bool = False,
-    extra_config: Optional[dict] = None,
-    extra_deps: Optional[Iterable[str]] = None,
```
pokey: The only use case I ever had for `extra_deps` was requiring the active bundle, but turns out that's done automatically by `active_bundle` so I don't have use for it anymore.  Client services are free to define this however they want though

microcosm_sagemaker/tests/bundles/compound.py
```diff
@@ -20,13 +20,15 @@ def dependencies(self) -> List[Bundle]:
         ]
 
     def fit(self, input_data: InputData) -> None:
-        pass
+        self.trained_param = self.simple_bundle.simple_trained_param + 1.0
```
pokey: This tests that the compound bundle can successfully access trained params of its dependencies

microcosm_sagemaker/tests/bundles/test_simple_bundle.py
```diff
@@ -1,57 +1,35 @@
-import tempfile
```
pokey: Most of this file has now been factored into the common `BundleTestCase` base class

microcosm_sagemaker/tests/test_dependency_traverser.py
```diff
@@ -13,13 +13,10 @@
...
 
-@dataclass
+@dataclass(eq=True, frozen=True)
 class Component:
     name: str
-    dependencies: List["Component"]
-
-    def __hash__(self):
-        return hash(self.name)
+    dependencies: List["Component"] = field(compare=False)
```
pokey: Semantically the same but leverages some `dataclasses` features to be a bit semantically richer

microcosm_sagemaker/testing/bundle.py
```diff
@@ -0,0 +1,154 @@
...
+    root_input_artifact_path: Path
+
+    @property
+    def extra_config(self) -> dict:
+        """
+        Derived classes can override this to provide extra config to the
+        various create_app functions.
+
+        """
+        return {}
```
pokey: I added this functionality because I see from @salmanmashayekh's [link](https://github.com/globality-corp/ml-nlp-serviceoffering-classifier/blob/master/ml_nlp_serviceoffering_classifier/tests/bundles/test_service_offering_bundle.py#L12) that he requires this for his use case.  Also seems helpful to allow for testing of bundle under different configurations.

PR #43, opened 2019-06-05
Support bundles which don't train

pokey:
Testing bundles which simply wrap other bundles and don't store any of their own artifacts was challening  with the new dependency setup due to the fact that git ignores empty directories.  This PR adds support for ignoring empty directories and adds a trivial bundle wrapper to test the functionality.

Also adds some testing for `directory_comparison`


pokey:
Superceded by #44

PR #44, opened 2019-06-05
Add support for bundles with empty artifact dir

salmanmashayekh:
* Add `ignore_hidden` to `directory_comparison`
* Add several unittests for directory comparison
* Add a new `noop_compound_bundle` bundle without any artifacts to the tests


microcosm_sagemaker/artifact.py
```diff
@@ -16,6 +16,9 @@ def __init__(self, path: Union[str, Path]):
     def init(self) -> None:
         self.path.mkdir(parents=True, exist_ok=True)
 
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({str(self.path)})"
```
pokey: I know I wrote this line, but on seeing it again I think we should wrap the path in quotes so that it’s a more accurate representation

microcosm_sagemaker/testing/directory_comparison.py
```diff
@@ -32,10 +33,12 @@ def directory_comparison(
     actual_paths = sorted([
         subpath.relative_to(actual_dir)
         for subpath in actual_dir.glob('**/*')
+        if not (subpath.name.startswith(".") and ignore_hidden)  # exclude hidden files if ignore_hidden is True
```
pokey: I’d make an `_is_hidden` function

microcosm_sagemaker/testing/directory_comparison.py
```diff
@@ -19,6 +19,7 @@ def directory_comparison(
     gold_dir: Path,
     actual_dir: Path,
     matchers: Optional[Mapping[Path, ExtractorMatcherPair]] = None,
+    ignore_hidden: bool = True,
 ):
     """
     Recursively checks the contents of `actual_dir` against the expected
```
pokey: My PR had added a check that the input directories exist because if not, then the test actually silently passes. This is bad because someone could get the paths wrong and think their test passes. Would be worth adding a test for this corner case (pass in nonexistent paths)

microcosm_sagemaker/tests/fixtures/directory_comparison.py
```diff
@@ -0,0 +1,203 @@
...
+            "subdir2": Dir({
+                ".keep": File(""),
+            }),
+        }),
+        actual=Dir({
+            "subdir1": Dir({
+                "hi.txt": File("hi"),
+            }),
+            "hello.txt": File("hello"),
+            "subdir2": Dir(),
```
pokey: It would be worth having a doubly-nested directory containing a hidden file

microcosm_sagemaker/tests/fixtures/directory_comparison.py
```diff
@@ -0,0 +1,203 @@
...
+
+    # Using ignore_hidden=True
+    DirectoryComparisonTest(
+        gold=Dir({
+            "subdir1": Dir({
+                "hi.txt": File("hi"),
+            }),
+            "hello.txt": File("hello"),
+            "subdir2": Dir({
+                ".keep": File(""),
```
pokey: Could just do no-args `File()`

microcosm_sagemaker/artifact.py
salmanmashayekh: like this @pokey ?
`f'{type(self).__name__}("{str(self.path)}")'`

microcosm_sagemaker/testing/directory_comparison.py
salmanmashayekh: Good catch! I added `test_nonexistent_directory_comparison` to `test_directory_comparison.py`.

microcosm_sagemaker/testing/bundle.py
```diff
@@ -128,8 +128,6 @@ def test_save(self) -> None:
 
 
 class BundleLoadTestCase(BundleTestCase, BundlePredictionChecker):
-    gold_bundle_output_artifact_path: Path
-
```
pokey: This wasn't necessary; I had removed in my original PR

microcosm_sagemaker/testing/directory_comparison.py
```diff
@@ -11,11 +11,11 @@
 from microcosm_sagemaker.testing.bytes_extractor import ExtractorMatcherPair
 
 
-def identity(x):
+def _identity(x):
     return x
 
 
-def is_hidden(path: Path) -> bool:
+def _is_hidden(path: Path) -> bool:
```
pokey: Until we have use for these in other files, better to mark them as ["hidden"](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles)

microcosm_sagemaker/testing/directory_comparison.py
```diff
@@ -31,21 +31,27 @@ def directory_comparison(
...
 
+    By default, this function ignores hidden files.  This functionality is
+    useful when you expect an empty directory, because git won't allow checking
+    in an empty directory.  In this situation you can add an empty `.keep` file
+    to the directory to make sure it is checked in.
+
     """
-    matchers = matchers or dict()
+    if matchers is None:
+        matchers = dict()
```
pokey: The above is more compact, but this is the [recommended](https://docs.python.org/3/faq/programming.html#why-are-default-values-shared-between-objects) idiom, due to the fact that the user could pass in an empty custom dict type

microcosm_sagemaker/testing/directory_comparison.py
```diff
@@ -31,21 +31,27 @@ def directory_comparison(
...
 
     assert_that(gold_dir.exists(), is_(True))
     assert_that(actual_dir.exists(), is_(True))
 
     actual_paths = sorted([
         subpath.relative_to(actual_dir)
         for subpath in actual_dir.glob('**/*')
-        if not (is_hidden(subpath) and ignore_hidden)  # exclude hidden files if ignore_hidden is True
+        if not (ignore_hidden and _is_hidden(subpath))  # exclude hidden files if ignore_hidden is True
```
pokey: I'd put `ignore_hidden` first because

- it is a bit more readable
- it is a bit more efficient when `ignore_hidden` is `False`, due to short-circuiting.  Not that that really matters at all here but no reason to [prematurely pessimize](https://www.oreilly.com/library/view/c-coding-standards/0321113586/ch10.html) 🤓

microcosm_sagemaker/tests/bundles/noop_compound.py
```diff
@@ -12,6 +12,7 @@
 class NoopCompoundBundle(Bundle):
     """
     This doesn't do any of its own training.
+
```
pokey: We always put a blank line before the end of docstrings

microcosm_sagemaker/tests/directory_test_description.py
```diff
@@ -13,28 +15,20 @@ def instantiate(self, path: Path) -> None:
         ...
 
 
+@dataclass
 class Dir(Entry):
-    def __init__(self, entries: Optional[Mapping[str, Entry]] = None):
-        if entries is None:
-            entries = dict()
-        self.entries = entries
+    entries: Mapping[str, Entry] = field(default_factory=dict)
```
pokey: Had the thought that a `dataclass` was cleaner here because it is a semantically richer way to capture all the behaviour I was defining manually

microcosm_sagemaker/tests/fixtures/directory_comparison.py
```diff
@@ -215,7 +214,6 @@ class DirectoryComparisonTest(NamedTuple):
         }),
         actual=Dir({
             "subdir1": Dir({
-                "hi.txt": File("hi"),
```
pokey: I wanted a doubly-nested totally empty dir, because that actually would have broken my version

microcosm_sagemaker/testing/directory_comparison.py
salmanmashayekh: Interesting! I did not know Python is smart enough to stop if the first condition of an AND is False! :D

PR #46, opened 2019-06-11
GLOB-33202 Fix runserver / wsgi parity

pokey:
Removes all setup code from `runserver` and folds it into the graph creation / configuration setup.  This step allows the `wsgi` code in our services to work automatically.  

Primarily, this change entails using a pipeline loader to rehydrate config from input artifacts, and creating a factory that will automatically load the active bundle from the input artifact.

Requires globality-corp/microcosm#41


microcosm_sagemaker/commands/evaluate.py
```diff
@@ -16,26 +16,20 @@
...
     graph: ObjectGraph,
     input_data: InputData,
-    root_input_artifact: RootInputArtifact,
 ) -> None:
-    # Load the saved artifact
-    graph.load_bundle_and_dependencies(
-        bundle=graph.active_bundle,
-        root_input_artifact=root_input_artifact,
-    )
-
```
pokey: This is now automatically done by `graph.use("load_active_bundle_and_dependencies")`

microcosm_sagemaker/loaders.py
```diff
@@ -36,59 +40,53 @@ def load_from_hyperparameters(metadata: Metadata) -> Configuration:
         return Configuration()
 
 
-def load_from_s3(url: str) -> Configuration:
+def s3_loader(metadata: Metadata, configuration: Configuration) -> Configuration:
```
pokey: This function now lets `two_stage_loader` automatically merge with input config.  Note that this means we now keep `"base_configuration"`, but this seemed like a minor cost for reducing the cognitive burden of having the merge code in here

PR #47, opened 2019-06-13
Release 0.4.0

pokey:


.circleci/config.yml
```diff
@@ -144,7 +144,12 @@ jobs:
             version=$(cat .bumpversion.cfg | awk '/current_version / {print $3}')
             python setup.py register -r pypi
             python setup.py sdist
-            twine upload dist/microcosm-sagemaker-${version}.tar.gz
+            twine upload --repository pypi dist/microcosm-sagemaker-${version}.tar.gz
+            version=$(python setup.py --version)
+            globality-build dependency-update \
+                --package-name microcosm_sagemaker \
```
KensoDev: This should be `microcosm-sagemaker` actually.

.globality/build.json
```diff
@@ -1,11 +1,14 @@
 {
   "params": {
+    "commitbot": {
```
KensoDev: If you enable commitbot, you should also add the permission for it in `terraform-github`.

.circleci/config.yml
pokey: Should this use `{{ name }}` then?  Or should it be something like

```py
{{ inflection.dasherize(import_name) }}
```

.globality/build.json
pokey: globality-corp/terraform-github#31

PR #58, opened 2019-07-15
Add `--profile`/`--no-profile` arguments to the `runserver` command

salmanmashayekh:


microcosm_sagemaker/profiling.py
```diff
@@ -0,0 +1,49 @@
...
+
+def enable_profiling(graph):
+    if "pyinstrument.profiler" not in modules:
+        raise Exception("Profiling requires 'pyinstrument'. Make sure it is installed.")
+
+    profile_dir = graph.config.flask.profile_dir or default_profile_dir(name=graph.metadata.name)
+    if not exists(profile_dir):
+        makedirs(profile_dir)
+
+    graph.app.config["PROFILE"] = True
```
piercefreeman: Is this required for profiling to be enabled?

salmanmashayekh: No, I think it is just a flag that shows it has been enabled. I am borrowing this from `microcosm-flask` @piercefreeman.

salmanmashayekh: I added a few lines to the README as you suggested @piercefreeman.

PR #65, opened 2019-07-29
Change the order of loaders in `train_conventions_loader`

salmanmashayekh:


pokey:
Can you please add test?

microcosm_sagemaker/tests/test_loaders.py
```diff
@@ -60,6 +60,32 @@ def test_train_conventions_loader(self):
...
+                loader = train_conventions_loader(
+                    initial_loader=load_from_dict(initial_configuration),
+                )
+                config = loader(metadata)
+
+        expected_config = dict()
+        for _config in [initial_configuration, hyperparameters, remote_configuration]:
+            expected_config.update(**_config)
+
+        assert_that(config, is_(equal_to(expected_config)))
```
pokey: I think it would make the test clearer to just check for the specific value you're after

pokey: better to use `has_entries` and put them directly into this line

PR #66, opened 2019-08-05
Add the `log_bundle_methods` decorator

salmanmashayekh:


microcosm_sagemaker/decorators.py
```diff
@@ -17,3 +19,64 @@ def factory(graph):
...
+        )
+
+
+class Timer:
+    def __enter__(self):
+        self._start = time.perf_counter()
+        return self
+
+    def __exit__(self, *args):
+        self.elapsed_seconds = time.perf_counter() - self._start
```
pokey: Turns out we already have this [in `microcosom-logging`](https://github.com/globality-corp/microcosm-logging/blob/9153bc4a05423158b5453c584ed4891adf5df9ae/microcosm_logging/timing.py), so probably better to use that.  However, I like our impl better because 
1. It uses [`time.perf_counter`](https://docs.python.org/3/library/time.html#time.perf_counter), and
2. I think using an object here is cleaner
I think it's worth:
1. Switching to using the existing `microcosm-logging` utility
2. Changing `microcosm-logging` to use `time.perf_counter` instead of `time`.  You don't need to wait on a release for that; not critical

I think the `microcosm-logging` interface is probably fine, so I don't think it's worth changing to our `class`-based approach

Make sense?  cc/ @GabeLambda

microcosm_sagemaker/decorators.py
```diff
@@ -17,3 +19,64 @@ def factory(graph):
...
+
+
+def _method_with_logging(original_method):
+    def new_method(*args, **kwargs):
```
pokey:
```suggestion
    def new_method(self, *args, **kwargs):
```

microcosm_sagemaker/decorators.py
```diff
@@ -17,3 +19,64 @@ def factory(graph):
...
+        self = args[0]
+        logging.info(
+            f"Started method `{original_method.__name__}` of the `{self.bundle_name}`."
+        )
+        with Timer() as t:
+            original_method(*args, **kwargs)
```
pokey:
```suggestion
            original_method(self, *args, **kwargs)
```

microcosm_sagemaker/decorators.py
```diff
@@ -17,3 +19,64 @@ def factory(graph):
...
+    def __init__(self, graph: ObjectGraph, **kwargs) -> None:
+        _init(self, graph, **kwargs)
+        self._graph = graph
+
+    @property
+    def bundle_name(self):
+        return get_component_name(self._graph, self)
+
+    cls.__init__ = __init__
+    cls.bundle_name = bundle_name
```
pokey: I could go either way on patching this onto the `cls`.  Could be useful to have around, but could also see trying to keep the class unaltered as much as possible

microcosm_sagemaker/decorators.py
```diff
@@ -17,3 +19,64 @@ def factory(graph):
             return component
         return factory
     return decorator
+
+
+# TODO: remove this when get_component_name is released with microcosm
+try:
```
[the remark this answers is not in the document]
salmanmashayekh: Yes, this is just a placeholder. I will release `microcosm`, fix this, and then merge it.

salmanmashayekh:
@pokey I update you about the `StopIteration` error on Slack. It is resolved now. The problem was that I was calling the `get_component_name` within the `__init__` of the class, hence, the object was not instantiated yet and could not be found.

microcosm_sagemaker/decorators.py
salmanmashayekh: I removed it to keep the class as intact as possible.

microcosm_sagemaker/decorators.py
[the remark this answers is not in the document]
salmanmashayekh: I removed it for clarity.

microcosm_sagemaker/decorators.py
salmanmashayekh: Makes sense, you are suggesting to keep the `elasped_time` in `microcosm-logging` a function, but use `perf_counter()` instead of `time()`.
What I am not familiar with is the advantage of  `perf_counter()` over `time()` @pokey. If you elaborate on that, I can create the `microcosm-logging` PR with the justification.

salmanmashayekh: Here is the feature branch: 
https://github.com/globality-corp/microcosm-logging/blob/feature/use-perf_counter-for-timing/microcosm_logging/timing.py

pokey: @salmanmashayekh the difference is that `time.time` could jump around due to daylight savings, changing clock time, etc.  `perf_counter` is designed for measuring time intervals

PR #67, opened 2019-08-17
Feature/add api health support to microcosm sagemaker

stephensxuglobality:
https://globality.atlassian.net/browse/DEVOPS-353

microcosm-sagemaker is currently using /ping as the health check since this is what SageMaker serve convention is expecting.

Our ECS conventions says that every service health is checked with /api/health and not /ping. We don’t want to hack the health check to go to /ping since our values are that all services should look the same.

All that being said, we want to add /api/health support to microcosm-sagemaker, that should live side-by-side with the /ping until we deprecate serving with SageMaker.

### DIFF
--- a/microcosm_sagemaker/conventions/health.py
+++ b/microcosm_sagemaker/conventions/health.py
@@ -1,0 +1,49 @@
+"""
+health controller.
+"""
+from marshmallow import Schema, fields
+from microcosm_flask.conventions.base import Convention
+from microcosm_flask.conventions.encoding import dump_response_data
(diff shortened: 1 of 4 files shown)


pokey:
@stephensxuglobality this is just temporary right?  I think the long-term solution is to switch our base uri to `/api` like our other services.

stephensxuglobality:
@pokey yes this is a temp solution, both @KensoDev and @piercefreeman have expressed same opinion that once we are off sagemaker, all routes should come back to the `/api` convention

PR #72, opened 2019-08-30
Add AllenNLP; migrate to functionality-based testing

pokey:


microcosm_sagemaker/bundle.py
```diff
@@ -24,21 +24,33 @@ def dependencies(self) -> List["Bundle"]:
...
         ...
 
-    @abstractmethod
     def fit(self, input_data: InputData) -> None:
         """
         Perform training
 
         """
-        ...
+        raise NotImplementedError
```
pokey: We now provide the option to either override `fit` and `save` or to just override `fit_and_save`.  The latter was needed because with AllenNLP it is not possible to separate fit from save

microcosm_sagemaker/random.py
```diff
@@ -14,4 +14,26 @@ def __init__(self, graph):
...
+            manual_seed(self.seed)
+        except ImportError:
+            pass
+
+        # Seed tensorflow if installed
+        try:
+            from tf.random import set_random_seed
+            set_random_seed(self.seed)
+        except ImportError:
+            pass
```
pokey: Seemed simplest to just automatically seed these if they're installed rather than relying on programmer to specify initializers manually

microcosm_sagemaker/bundle.py
piercefreeman: @pokey True if we use `train_model` from `allennlp.commands.train`.  That said, we should be able to re-implement our own and keep the distinction between `fit` and `save`.

I realize doing so may result in us not getting the latest and greatest from allennlp if they update their training pipeline.  But I think it would be worth it to keep the abstraction at fit/save instead of having to develop a new function that's only used in special cases.  WDYT?

microcosm_sagemaker/pickling_bundle.py
```diff
@@ -0,0 +1,38 @@
+from typing import List
+
+import joblib
+
+from microcosm_sagemaker.artifact import BundleInputArtifact, BundleOutputArtifact
+from microcosm_sagemaker.bundle import Bundle
+
+
+class PicklingBundle(Bundle):
```
piercefreeman: Is this a common enough bundle to add to the `microcosm-sagemaker` root repo?  It looks like we're just using it right now as part of unit tests... Can we isolate there?

microcosm_sagemaker/testing/bundle.py
```diff
@@ -152,3 +141,97 @@ def test_load(self) -> None:
         )
 
         self.check_bundle_prediction(self.graph.active_bundle)
+
+
+class BundleFitSaveLoadTestCase(BundleTestCase, BundlePredictionChecker):
+    """
+    One combined test of fit, save and load.  First fits and saves bundle, then
+    optionally tests artifact against a gold directory, then loads the trained
```
piercefreeman: Any reason to keep the gold directory around?  Any time the model changes we have to regenerate the contents of it, so its ability to sanity check model results is a bit limited.

And the functional model tests that you introduce should do away with this need completely, no?

microcosm_sagemaker/frameworks/allennlp/bundle.py
```diff
@@ -0,0 +1,59 @@
...
+from allennlp.predictors.predictor import Predictor
+
+import microcosm_sagemaker.frameworks.allennlp.vanilla_predictor  # noqa
+from microcosm_sagemaker.artifact import BundleInputArtifact, BundleOutputArtifact
+from microcosm_sagemaker.bundle import Bundle
+from microcosm_sagemaker.frameworks.allennlp.constants import ARTIFACT_NAME, CUDA_DEVICE
+from microcosm_sagemaker.input_data import InputData
+
+
+class AllenNLPBundle(Bundle):
```
salmanmashayekh: Why are we adding `AllenNLP` to `microcosm-sagemaker`? Is there a plan to add more frameworks, as well @pokey?

microcosm_sagemaker/bundle.py
pokey: @piercefreeman your proposal here was definitely my first choice, but after sinking some time into trying to figure out how to split up the steps in `allennlp`, I decided that the cost of supporting both abstractions was not great enough to warrant the effort.  If you know of an easy way to do it I'm all ears.  cc/ @salmanmashayekh @andrewplumb / whoever else is familiar with Allennlp

microcosm_sagemaker/pickling_bundle.py
pokey: See https://github.com/globality-corp/ml-taxonomysimilarity-classifier/pull/49#discussion_r319896593.  It's only 1 use case, so maybe a bit premature, but I believe I've seen us doing this elsewhere as well.  Happy to fold back into `taxonomy-similarity` if you think it's too early to move into `microcosm-sagemaker`

microcosm_sagemaker/testing/bundle.py
pokey: Mostly keeping for backward compatibility.  Once we've removed all usage we could probably deprecate gold functionality.  I do think it's useful to support a certain amount of directory checking to make sure undesired files aren't creeping into our artifacts, but we can build that out as needed

microcosm_sagemaker/frameworks/allennlp/bundle.py
pokey: See https://github.com/globality-corp/ml-taxonomysimilarity-classifier/pull/49#discussion_r319896936.  And yes, if we see useful abstractions there we'd like to support more frameworks

microcosm_sagemaker/bundle.py
piercefreeman: @pokey Both abstractions being the current ones for `fit`/`save`?  The underlying `Trainers` that manage the process take some init params which control whether they cache intermediate results, which could partially address this:

https://github.com/allenai/allennlp/blob/27ebcf6ba3e02afe341a5e62cb1a7d5c6906c0c9/allennlp/training/trainer.py#L108

That said - I think this may be a bigger question.  I can see a clear use-case for a custom fitting function wanting to cache intermediary results to disk so it can keep track of the best model (versus carrying it alongside the current parameters in memory).  Should that be allowed?  Would we want to enforce the `fit_save` function in that case since they are writing intermediate files to disk?  Or would we prefer the user to establish a temp directory to store these files during fitting, and only copy them to a permanent location during save?

microcosm_sagemaker/pickling_bundle.py
piercefreeman: @pokey For the time-being, I think let's move it back.  If we notice ourselves doing it in a lot of different models than we can reconsider.  But tentatively I'd say that we should keep `microcosm_sagemaker` free of business logic bundles and encourage a common bundle library if we really see ourselves using them over and over.

microcosm_sagemaker/bundle.py
pokey: Both abstractions meaning supporting both of
- Today's `fit` / `save` api
- The new `fit_and_save` api defined in this PR

The issue of intermediate files is related, and is something we need to revisit wrt allennlp in particular.  today, our output artifact for `taxonomy-similarity` includes all logs and intermediate model files, which blows up the artifact size by 10-50x.  This update doesn't address that problem.  I've filed [GLOB-36015](https://globality.atlassian.net/browse/GLOB-36015) for that, but I don't think we need to solve it before we merge this PR.

microcosm_sagemaker/pickling_bundle.py
pokey: 3 more examples across 2 other repos:
- https://github.com/globality-corp/ml-projectcasesimilarity-classifier/blob/fe5e42826df4afac4078dfd61ac3335213bab2e1/ml_projectcasesimilarity_classifier/bundles/text_preprocessor_bundle.py#L162-L183
- https://github.com/globality-corp/ml-projectcasesimilarity-classifier/blob/fe5e42826df4afac4078dfd61ac3335213bab2e1/ml_projectcasesimilarity_classifier/bundles/tfidf_vectorizer_bundle.py#L78-L86
- https://github.com/globality-corp/ml-nlp-qna-choice-suggestion/blob/68b14e9a9eb0ad30e6691d7741a28831279705a7/ml_nlp_qna_choice_suggestion/bundles/inverse_doc_freq_bundle.py#L40-L48

still unconvinced?

pokey: Resolving this based on phone conversation where @piercefreeman and I agreed to leave it in `microcosm-sagemaker` library

microcosm_sagemaker/bundle.py
```diff
@@ -24,33 +24,21 @@ def dependencies(self) -> List["Bundle"]:
...
         ...
 
+    @abstractmethod
     def fit(self, input_data: InputData) -> None:
         """
         Perform training
 
         """
-        raise NotImplementedError
+        ...
```
piercefreeman: @pokey Is this a new python syntax for `pass`?

pokey: https://github.com/python/typing/issues/109

PR #75, opened 2019-09-16
Fix Allennlp move operation

pokey:
The allennlp copy operation was failing in SageMaker due to the fact that the tmp directory is in a different filesystem than the output artifact directory

### DIFF
--- a/microcosm_sagemaker/frameworks/allennlp/bundle.py
+++ b/microcosm_sagemaker/frameworks/allennlp/bundle.py
@@ -45,6 +45,10 @@ class AllenNLPBundle(Bundle):
         allen_nlp_path.mkdir(parents=True)
 
         for child in self.temporary_allennlp_path.iterdir():
+            # NB: It is important to use `shutil.move` here rather than
+            # `rename` because the latter works across filesystems, and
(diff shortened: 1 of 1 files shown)


microcosm_sagemaker/frameworks/allennlp/bundle.py
```diff
@@ -44,7 +45,11 @@ def save(self, output_artifact: BundleOutputArtifact) -> None:
         allen_nlp_path.mkdir(parents=True)
 
         for child in self.temporary_allennlp_path.iterdir():
-            child.rename(allen_nlp_path / child.name)
+            # NB: It is important to use `shutil.move` here rather than
+            # `rename` because the former works across filesystems, and
+            # SageMaker uses a different filesystem for its temporary
+            # directories and the output artifact directory
+            shutil.move(child, allen_nlp_path / child.name)
```
adamhadani: any unit-tests worth adding around this use case?

pokey: The problem is that it only happens when you're trying to move a file from one filesystem to another, which I'm not sure how to arrange for during unit testing

PR #82, opened 2019-09-24
Use self.logger for auto logging

pokey:
Some minor improvements to automatic bundle logging:
- Use `self.logger`
- Use `g` instead of `f` in format string

### DIFF
--- a/microcosm_sagemaker/decorators.py
+++ b/microcosm_sagemaker/decorators.py
@@ -1,7 +1,5 @@
-import logging
 from typing import Any, Callable
 
-from microcosm.api import get_component_name
 from microcosm.object_graph import ObjectGraph
 from microcosm_logging.timing import elapsed_time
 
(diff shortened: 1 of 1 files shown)


microcosm_sagemaker/decorators.py
```diff
@@ -24,16 +22,13 @@ def factory(graph):
...
+        self.logger.info(f"Started method `{original_method.__name__}`.")
         timing = {}
         with elapsed_time(timing):
             original_method(self, *args, **kwargs)
-        logging.info(
-            f"Completed method `{original_method.__name__}` of the `{bundle_name}` "
-            f"after {timing['elapsed_time']/1000:.1f} seconds."
+        self.logger.info(
+            f"Completed method `{original_method.__name__}`"
+            f"after {timing['elapsed_time']/1000:.3g} seconds."
```
pokey: This format uses sig figs instead of fixed-point, so it still works well for very small and very large times

microcosm_sagemaker/decorators.py
```diff
@@ -24,16 +22,13 @@ def factory(graph):
 
 def _method_with_logging(original_method):
     def new_method(self, *args, **kwargs):
-        bundle_name = get_component_name(self._graph, self)
-        logging.info(
-            f"Started method `{original_method.__name__}` of the `{bundle_name}`."
-        )
+        self.logger.info(f"Started method `{original_method.__name__}`.")
```
pokey: Using `self.logger` means it doesn't go onto root logger, and then we don't need to manually specify bundle name

PR #86, opened 2019-11-06
Weights & Biases Integration

pokey:
An example usage of the new metric logging:
https://github.com/globality-corp/qna-choice-suggestion/pull/7

An example experiment:
https://app.wandb.ai/globality/qna-choice-suggestion/runs/zwmwmf7l/overview
_note_: we are only logging a `run_summary` in this experiment.


microcosm_sagemaker/metrics.py
```diff
@@ -0,0 +1,19 @@
...
+        self.metrics_observers = []
+
+    def register(self, observer):
+        self.metrics_observers.append(observer)
+
+    def log_metric(...):
+        for metric_observer in metric_observers:
+            metric_observer.log_metric(...)
+
+graph.metrics.log_metric(...)
```
pokey: this is usage: delete here

microcosm_sagemaker/wandb.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@training_initializer()
+@metrics_observer()
+class WeightsAndBiases:
+    def __init__(self, graph):
+        pass
+
+    def init(self):
+        wandb.init()
```
pokey: fill this function out with other `wandb` init stuff

microcosm_sagemaker/wandb.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def log_metric(...):
+        wandb.log_metric(...)
+
+graph.use(
+    ...
+    "wandb",
+)
```
pokey: more usage example: remove

microcosm_sagemaker/metrics.py
```diff
@@ -0,0 +1,19 @@
...
+
+    """
+    def __init__(self, graph):
+        self.graph = graph
+        self.metrics_observers = []
+
+    def register(self, observer):
+        self.metrics_observers.append(observer)
+
+    def log_metric(...):
```
pokey: maybe change the param signature here and add other log functions mirroring wandb api

microcosm_sagemaker/decorators.py
```diff
@@ -20,6 +20,22 @@ def factory(graph):
     return decorator
 
 
+def metrics_observer():
+    """
+    Register a microcosm component as a training initializer, so that its init
+    method will automatically be called.  This function is designed to be used
+    as a decorator on a factory.
```
pokey: Update docstring

microcosm_sagemaker/metric_loggers.py
```diff
@@ -0,0 +1,21 @@
+class MetricLoggers:
+    """
+    The training initializer registry is a place to register functions that need to be
+    called during training initialization.  A typical example of this is to
+    seed random number generators.
```
pokey: Update docstring

microcosm_sagemaker/metrics/store.py
```diff
@@ -39,8 +39,8 @@ def create(self, model_name, metrics=[], hyperparameters={}, mode=MetricMode.TRA
                 MetricData=metric_data,
```
pokey: How hard would it be to rig this up to use the new metrics setup?

microcosm_sagemaker/metrics/store.py
```diff
@@ -39,8 +39,8 @@ def create(self, model_name, metrics=[], hyperparameters={}, mode=MetricMode.TRA
...
             )
         except (ClientError, NoCredentialsError, NoRegionError):
-            print("CloudWatch publishing disabled")  # noqa: T003
-            print(dumps(metric_data, indent=4))  # noqa: T003
+            print("CloudWatch publishing disabled")  # noqa: T001
+            print(dumps(metric_data, indent=4))  # noqa: T001
```
pokey: Better to use logger than print, then we can remove these noqa's and it will automatically go to cloudwatch etc

microcosm_sagemaker/mixins/metric.py
```diff
@@ -43,7 +43,7 @@ def log_best_stats(self, key, mode, best_fn=max):
         # Get the best metric
         best_metric = self.get_best_metric(key, mode, best_fn)
 
-        print(f"Best Epoch ({key}, {mode}) = {best_metric.epoch}")  # noqa: N123
+        print(f"Best Epoch ({key}, {mode}) = {best_metric.epoch}")  # noqa: T001
```
pokey: as above

microcosm_sagemaker/wandb.py
```diff
@@ -0,0 +1,33 @@
...
+@training_initializer()
+@metrics_observer()
+class WeightsAndBiases:
+    def __init__(self, graph):
+        self.graph_config = graph.config
+        self.testing = graph.metadata.testing
+        self.name = graph.metadata.name
+
+    def init(self):
+        # os.environ["WANDB_API_KEY"] = "XXXXXXXX"
```
pokey: Is this just to track the DEVOPS request to get this secret?

microcosm_sagemaker/decorators.py
```diff
@@ -47,3 +63,16 @@ def __init__(self, graph: ObjectGraph, **kwargs) -> None:
     cls.save = _method_with_logging(cls.save)
 
     return cls
+
+
+def metrics(cls):
```
pokey: Not sure what it buys us to have a decorator vs just `graph.metrics`.  I could see this being useful if we did something like we do with `@logger` where we use the class name to create a log context

microcosm_sagemaker/metric_loggers.py
```diff
@@ -0,0 +1,21 @@
+class MetricLoggers:
```
pokey: Maybe just `Metrics`?

microcosm_sagemaker/metrics/store.py
salmanmashayekh: This is from before, I only made sure it passes `linting`. I assumed there was a reason we were printing instead of logging. 
@piercefreeman ?

microcosm_sagemaker/wandb.py
salmanmashayekh: Yes.

microcosm_sagemaker/decorators.py
salmanmashayekh: Well, I thought it is slicker. Because the other approach requires every user class to add `self.metrics=graph.metrics` to the `__init__`, and the decorator approach does not. But I am okay either way...

microcosm_sagemaker/metric_loggers.py
salmanmashayekh: The class name can be `Metrics` or `MetricLoggers`, does not matter. But `metrics` cannot be attached to the graph, since it gets into some sort of conflict with another `metrics`, I think coming from `microcosm_flask`. So for consistency, I used `metric_loggers` for filename, class name, and binding name.

microcosm_sagemaker/decorators.py
pokey: I think I'd favour explicit over implicit unless we have good reason to use the decorator.  If we do add context based on class where it is called as we do with `@logger`, then I could see using the decorator.  Otherwise I think it is clearer to use the normal pattern of `self.foo = graph.foo`.  Plus then we don't need the `mypy` `logger` attribute [hack](https://globality.atlassian.net/wiki/display/GLOB/customcontent/928484515).

microcosm_sagemaker/metric_loggers.py
pokey: Hmm right.  Maybe worth a bit of thought about naming here.  `experiment_metrics`? cc/ @piercefreeman

microcosm_sagemaker/decorators.py
salmanmashayekh: The `metrics` decorator removed

microcosm_sagemaker/metrics/store.py
salmanmashayekh: I will give it a shot!

microcosm_sagemaker/metrics/metric_loggers.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, graph):
+        self.graph = graph
+        self.testing = graph.metadata.testing
+        self.metric_observers = []
+
+    def register(self, observer):
+        if not self.testing:
+            self.metric_observers.append(observer)
+
+    def log_time_series(self, *args, **kwargs):
```
salmanmashayekh: Per Pierce's recommendation, changed the names for better readability: `log_time_series` and `log_static`.

microcosm_sagemaker/metrics/sagemaker/store.py
```diff
@@ -0,0 +1,86 @@
...
+    for k, v in d.items():
+        new_key = f"{parent_key}{sep}{k}" if parent_key else k
+        if isinstance(v, collections.MutableMapping):
+            items.extend(flatten(v, new_key, sep=sep).items())
+        else:
+            items.append((new_key, v))
+    return dict(items)
+
+
+def normalize_hyperparam_value(v):
```
salmanmashayekh: Empty strings fail `botocore`'s parameter validation. What do you suggest I replace them with @pokey @piercefreeman?

microcosm_sagemaker/mixins/metric.py
```diff
@@ -1,3 +1,4 @@
+import logging
 from abc import ABCMeta
```
salmanmashayekh: As far as I can see, this mixin is not used in any repo. How about we just take it out?

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, graph):
+        self.graph_config = graph.config
+        self.testing = graph.metadata.testing
+        self.project_name = graph.metadata.name.replace("_", "-")
+
+    def init(self):
+        # Only initialize wandb if it is not a testing
+        if not self.testing:
+            # TODO: Remove this line if devops come up with a solution
+            environ["WANDB_API_KEY"] = "e9698d764c80c1076a1ff017fa9f722b3135624a"
```
salmanmashayekh: Everyone seems to be on the same page about `wandb` not hosting any sensitive data (I also talked to @adamhadani about it). So, I suggest we move forward with hard-coding the key in the repo and unblocking ourselves. If the devops team comes up with a solution, we can delete this key and generate a new one.

microcosm_sagemaker/metric_loggers.py
salmanmashayekh: I actually like `experiment_metrics` much better. Renamed!

microcosm_sagemaker/metrics/wandb/store.py
piercefreeman: @salmanmashayekh I'm okay with having the key in plaintext within each independent model, but this isn't the right repo for it.  `microcosm-sagemaker` is public - so even if we're not worried about sharing of secrets internally this is not something to do within the shared source.  We can set this at the level of the graph config for each model.

You should remove this line and regenerate your API key.

salmanmashayekh: Absolutely! I moved the key to `config.wandb.api_key` to be set by individual models. And deleted this key.

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, graph):
+        self.graph_config = graph.config
+        self.testing = graph.metadata.testing
+        self.project_name = graph.metadata.name.replace("_", "-")
+
+    def init(self):
+        # Only initialize wandb if it is not a testing
+        if not self.testing:
+            # TODO: Remove this line if devops come up with a solution
+            environ["WANDB_API_KEY"] = self.graph_config.wandb.api_key
```
pokey: Hmm.  I'd press devops a bit on this.  This setup is def not ideal.

microcosm_sagemaker/hyperparameters.py
```diff
@@ -0,0 +1,56 @@
...
+        ...
+    ```
+
+    """
+    if not parameter_type:
+        parameter_type = type(default_value)
+    return HyperParameter(type=parameter_type, default_value=default_value)
+
+
+def get_graph_hyperparams():
```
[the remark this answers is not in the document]
salmanmashayekh: Correct.
Also, to indicate important hyperparameters, i.e. parameters that can potentially be fine-tuned, for training.

microcosm_sagemaker/hyperparameters.py
```diff
@@ -0,0 +1,56 @@
...
+
+    hyperparams = []
+
+    for binding_name, class_definition in _registry.all.items():
+        if "_defaults" in class_definition.__dict__:
+            hyperparams.extend([
+                (binding_name, hyperparamer_name)
+                for hyperparamer_name, hyperparamer in class_definition.__dict__["_defaults"].items()
+                if getattr(hyperparamer, "is_hyperparameter", False)
+            ])
```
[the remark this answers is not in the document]
salmanmashayekh: OK, I addressed your suggestions in the latest PR @pokey. I have added a ton of comments to make it clear. Plz see the updated `hyperparameters.py` and `test_hyperparameters.py`.

microcosm_sagemaker/metrics/wandb/store.py
salmanmashayekh: I agree and I will try to meet @KensoDev tomorrow to discuss.

But I am still going to advocate for the existing non-ideal, but safe-enough, approach. The data is not sensitive, so we should not be blocked while waiting.

microcosm_sagemaker/metrics/experiment_metrics.py
```diff
@@ -0,0 +1,35 @@
+from microcosm_logging.decorators import logger
+
+
+@logger
+class ExperimentMetrics:
+    """
+    This registry is a place to register functions that log metrics.
+    This registry calls the `init` method of the registered components
```
pokey: One last docstring to update re `init`

microcosm_sagemaker/metrics/experiment_metrics.py
```diff
@@ -0,0 +1,35 @@
...
+    to initialize them.
+    The `log_metric` and `run_summary` methods are used to store multiple-value
```
pokey: Also `run_summary` is no longer a thing

microcosm_sagemaker/metrics/sagemaker/store.py
```diff
@@ -0,0 +1,97 @@
...
+from microcosm_logging.decorators import logger
+
+from microcosm_sagemaker.decorators import metrics_observer, training_initializer
+from microcosm_sagemaker.hyperparameters import get_hyperparameters
+from microcosm_sagemaker.metrics.sagemaker.models import LogMode, MetricUnit
+
+
+# NOTE: We can only use up to 10 dimensions:
+# https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudwatch.html
+MAX_DIMENSIONS = 10
```
pokey: Please put in `constants.py`

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -0,0 +1,46 @@
...
+        self.graph = graph
+        self.testing = graph.metadata.testing
+        self.project_name = graph.metadata.name.replace("_", "-")
+
+    def init(self):
+        # Only initialize wandb if it is not a testing
+        if not self.testing:
+            # TODO: Remove this line if devops come up with a solution
+            # https://globality.atlassian.net/browse/DEVOPS-635
+            environ["WANDB_API_KEY"] = self.graph.config.wandb.api_key
```
pokey: What is the advantage of setting the env var here vs just letting calling code set env var?  Is that because it's easier to put it in microcosm config sometimes?  Either way, might be slightly more usable to say
```suggestion
            if self.graph.config.wandb.api_key:
                environ["WANDB_API_KEY"] = self.graph.config.wandb.api_key
```
so that you don't overwrite it if they already put it there

microcosm_sagemaker/tests/metrics/bundle.py
```diff
@@ -0,0 +1,23 @@
...
+
+from microcosm_sagemaker.hyperparameters import hyperparameter
+
+
+@binding("bundle_with_metric")
+@defaults(
+    param=1,
+    hyperparam=hyperparameter(2),
+)
+class BundleWithMetirc():
```
pokey:
```suggestion
class BundleWithMetric():
```

microcosm_sagemaker/tests/metrics/test_cloudwatch.py
```diff
@@ -0,0 +1,83 @@
...
+class TestCloudwatch():
+    def setup(self):
+        self.graph = create_object_graph(
+            name="test-project",
+            testing=False,
+        )
+        self.graph.use("training_initializers", "cloudwatch", "bundle_with_metric")
+        self.graph.lock()
+
+        self.dimensions = [
```
pokey:
```suggestion
        self.expected_dimensions = [
```

microcosm_sagemaker/tests/metrics/test_cloudwatch.py
```diff
@@ -0,0 +1,83 @@
...
+)
+from microcosm.api import create_object_graph
+
+from microcosm_sagemaker.metrics.sagemaker.models import MetricUnit
+
+
+class TestCloudwatch():
+    def setup(self):
+        self.graph = create_object_graph(
+            name="test-project",
```
pokey: I'd do `self.name = "test-project"` above so that you don't need to repeat this string below

microcosm_sagemaker/metrics/wandb/store.py
salmanmashayekh: The calling code providers the API key, since it will be in the config of the ML models. But the _env var_ is set here in `microcosm-sagemaker`.

PR #90, opened 2019-11-15
Add documentation

salmanmashayekh:
GLOB-38182 Weights and Biases integration


README.md
```diff
@@ -18,4 +18,72 @@ To enable profiling of the app, use the `--profile` flag with `runserver`:
...
+
+```
+{
+    "wandb": {
+        "api_key": "XXXXXX"
+    }
+}
+```
+
+If you don't have access to `wandb` to create a new API key, contact @pierce to be invited.
```
pokey: one thing to keep in mind here is that this is a public repo, so not sure if we want to have this here cc/ @piercefreeman

PR #91, opened 2020-01-02
Fix the issue with the `enabled` flag for cloudwatch store

salmanmashayekh:
The `enabled` flag was misplaced. Instead of being under the `__init__` method, it was under the `init`.


salmanmashayekh:
We already have decent test coverage for the metrics. This one is hard to add and I would say not worth it.

PR #99, opened 2020-04-22
GLOB-44216 Add PipelineHarness for tests

marcglobality:
This task allows to have this in a centralised place, so that all ML repos can import from here

As a byproduct, I added a minor change to the original version. From Python3.3+  `@abstractproperty` becomes `@property + @abstractmethod`

Extras
------

I want to take this opportunity also to propose (and we could start with this repo) to have tests outside of the module. This obviously wouldn't include the file added in this PR.

This is recommended e.g. here:

* https://github.com/pypa/sampleproject
* https://docs.python-guide.org/writing/structure/ (20k stars)
* https://github.com/drivendata/cookiecutter-data-science


marcglobality:
> LGTM but please add a test or two for this functionality.
> 
> Re having tests outside the module, that’s a bigger discussion because our infrastructure assumes that in various places. I don’t disagree, but maybe raise on Slack and we can start discussion there and move to a tech design / ticket if there’s buy-in

Well it looks like it didn't have much success among engineers :D 

About this PR. Used this opportunity to simplify the code, make it more robust (`co_arguments` vs `signature().parameters`) and make the name of the class and file be the same. I would like to have `TestingPipeline` but I think anything starting with `test`, nose thinks it's a test and breaks so went back to `PipelineHarness`. I could have come up with something more convoluted like `LinearTestingChain`, just to include the word test, but I think it's just fine 😅

microcosm_sagemaker/testing/pipeline_harness.py
```diff
@@ -0,0 +1,26 @@
+from abc import ABC, abstractmethod
+from inspect import signature
+
+
+class PipelineHarness(ABC):
+
```
pokey, line 6: nit: we usually don't have an empty line here
```suggestion
```

microcosm_sagemaker/testing/pipeline_harness.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    Base class to support feature tests, by specifying the steps/pipes
+    of a pipeline
+    """
```
pokey, line 10: nit: We use an empty line before end of docstring
```suggestion

    """
```

microcosm_sagemaker/testing/pipeline_harness.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @property
+    @abstractmethod
+    def _steps(self):
```
pokey, line 14: Would be good to use type annotation here.  I think probably
```suggestion
    def _steps(self) -> Iterable[Callable[..., Optional[Mapping]]]:
```
Docstring would be good too

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
```diff
@@ -0,0 +1,82 @@
...
+            self.step_training,
+            self.step_prepare_predict,
+            self.step_predict,
+            self.remove_filesystem,
+        ]
+
+    def create_filesystem(self):
+        directory = TemporaryDirectory()
+
+        return {"artifact_directory": directory, "artifact_path": directory.name}
```
pokey, line 44: We prefer `dict` constructors to dict literals for keys that are snake-cased strings

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
```diff
@@ -0,0 +1,82 @@
...
+            root_output_artifact=RootOutputArtifact(artifact_path),
+        )
+
+    @mock_app_hooks()
+    def step_prepare_predict(self, artifact_path):
+        graph = create_serve_app(
+            extra_loader=load_from_dict(root_input_artifact_path=artifact_path,),
+            debug=True,
+        )
+        return {"graph": graph, "client": graph.flask.test_client()}
```
pokey, line 61: as above

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
```diff
@@ -0,0 +1,82 @@
...
+
+class TestSimpleBundlePipeline(PipelineHarness, TestCase):
+    config = {
+        "active_bundle": "simple_bundle",
+        "active_evaluation": "simple_evaluation",
+    }
+
+    # TODO: add more test cases
+    expected = [{"uri": "http://simple.com", "score": 3.0}]
+    test_cases = [{"input": {"simpleArg": 1.0}, "expected": expected}]
```
pokey, line 29: I'd use `dict` constructor and probably just fold `expected` in to the expression.  But as mentioned below I think this machinery is not necessary for this simple test case

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
```diff
@@ -0,0 +1,82 @@
...
+
+            expected = test_case["expected"]
+            assert_that(
+                len(response_items), is_(equal_to(len(expected))),
+            )
+            for response_item, expected_item in zip(response_items, expected):
+                assert_that(response_item["uri"], is_(equal_to(expected_item["uri"])))
+                assert_that(
+                    response_item["score"], is_(close_to(expected_item["score"], 1e-6))
+                )
```
pokey, line 79: I don't think all this machinery / abstraction buys us much.  I'd just do

```suggestion
    def step_predict(self, client):
        response = client.post(
            "/api/v1/invocations",
            json=dict(simpleArg=1.0),
        )

        assert_that(response.status_code, is_(equal_to(200)))
        assert_that(
            response.items,
            has_entries(
                uri="http://simple.com",
                score=close_to(3.0),
            ),
        )
```

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
```diff
@@ -0,0 +1,82 @@
...
+    def step_training(self, artifact_path):
+        graph = create_train_app(extra_loader=load_from_dict(self.config), testing=True)
+        run_train(
+            graph=graph,
+            input_data=InputData(get_fixture_path("input_data")),
+            root_output_artifact=RootOutputArtifact(artifact_path),
+        )
+
+    @mock_app_hooks()
+    def step_prepare_predict(self, artifact_path):
```
pokey, line 56: I think I'd lean towards `inference` instead of `predict`, because it is testing the full inference serving machinery rather than just the `bundle.predict` function

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
```diff
@@ -0,0 +1,82 @@
...
+        for test_case in self.test_cases:
+            response = client.post("/api/v1/invocations", json=test_case["input"])
+            response_items = response.json["items"]
+            assert_that(response.status_code, is_(equal_to(200)))
+
+            expected = test_case["expected"]
+            assert_that(
+                len(response_items), is_(equal_to(len(expected))),
+            )
+            for response_item, expected_item in zip(response_items, expected):
```
ml-globality, line 75: I agree with @pokey that this can be simplified

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
marcglobality: I blacked the code, let me know if you want it the way you formatted it

microcosm_sagemaker/testing/pipeline_harness.py
```diff
@@ -0,0 +1,26 @@
...
+        pass
+
+    def test_steps(self):
+        store = {}
```
ml-globality, line 18: nit: I rather liked the original variable name `step_input` -- it was a bit more specific whereas `store` is very generic, but not a big deal

marcglobality: I don't like `step_input` because it's misleading. This is saving *everything* we have, and then we filter / pick from the store to create the "step_input" in line 33

microcosm_sagemaker/tests/bundles/test_simple_bundle_pipeline.py
marcglobality: maybe prepare_serve vs prepare_inference  ? given that it’s doing create_serve_app

PR #107, opened 2020-07-09
Add the wandb run url to the landing page

salmanmashayekh:


microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -64,6 +64,12 @@ def _create_new_wandb_run(self, notes=None):
         # Injecting the wandb run path into the config
         self.graph.config.wandb.run_path = wandb_run.path
 
+        # Adding the link to the Weights & Biases run to the landing page
+        # https://github.com/globality-corp/microcosm-flask/blob/develop/microcosm_flask/conventions/landing.py#L68
```
pokey, line 68: I think I'd remove this link.  It will get stale given that it has a line number.  Ideally microcosm-flask would have documentation for this that you could link to

salmanmashayekh: I could not find anything in the README or in Confluence. Removing the link for now...

PR #113, opened 2020-11-25
GLOB-47640 Minor changes to the wandb metrics store

salmanmashayekh:
- The latest `wandb` version breaks the metric logging wrapper of `microcosm-sagemaker`. This patch will fix that issues.
- Also, it adds `training_initializers.init()` to `run_eval`, which makes it similar to `run_train`. 
- It also adds an `enable/disable` config option for the `wandb` metric, in order to disentangle its being on/off from the `metadata.testing`
- It rips out `allennlp` framework from `microcosm-sagemaker`.


microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -31,7 +39,7 @@ def init(self):
         os.environ["WANDB_ENTITY"] = "globality"
 
         # Only initialize wandb if it is not a testing
-        if self.testing:
+        if not self.enable:
```
marcglobality, line 42: should the comment be updated?

marcglobality (approved):
LGTM

microcosm_sagemaker/commands/evaluate.py
```diff
@@ -31,5 +31,7 @@ def run_evaluate(
     graph: ObjectGraph,
     input_data: InputData,
 ) -> None:
-    # Evaluate
+
+    graph.training_initializers.init()
```
pokey, line 35: It's a bit confusing to call these `training_initializers` if they also run during `evaluate`.  What's the rationale for running them here?

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -15,10 +16,17 @@
...
+@defaults(
+    enable=None,
+)
 class WeightsAndBiases:
     def __init__(self, graph):
         self.graph = graph
-        self.testing = graph.metadata.testing
+        self.enable = (
+            # If it is not explicitly enabled or disabled, enable if not testing.
+            graph.config.weights_and_biases.enable or not graph.metadata.testing
```
pokey, line 28: @ml-globality @tomaszpietruszka-globality I recall having a conversation with one of you about this, but I don't remember what the conclusion was

pokey (commented):
I think this one is fine but had a couple questions before I sign off

microcosm_sagemaker/commands/evaluate.py
salmanmashayekh: I agree the name is not very appropriate anymore. We call it during the evaluation to load the wandb run from the training, so that we can push evaluation metrics into the same run. 
Similar to what we do here for example: https://github.com/globality-corp/service-offering-classifier/blob/master/service_offering_classifier/tests/bundles/test_artifact_performance.py#L52

microcosm_sagemaker/metrics/wandb/store.py
salmanmashayekh: The idea was to have an on/off switch for wandb that works independently from the `testing` flag. And we are adding that switch here.

salmanmashayekh: Right @ml-globality ?

salmanmashayekh:
@pokey Do I have your blessing?! :D

microcosm_sagemaker/commands/evaluate.py
pokey: I think this is ok but maybe add something to a doc / doc string somewhere to indicate this potential gotcha

pokey (approved):

PR #114, opened 2020-12-02
Release/0.8.7

salmanmashayekh:
A few minor updates, the main one being compatibility with wandb >= 0.10.


microcosm_sagemaker/commands/evaluate.py
```diff
@@ -31,5 +31,7 @@ def run_evaluate(
     graph: ObjectGraph,
     input_data: InputData,
 ) -> None:
-    # Evaluate
+
+    graph.training_initializers.init()
```
piercefreeman, line 35: This seems like something we don't want at eval time, no?

microcosm_sagemaker/decorators.py
```diff
@@ -10,6 +10,9 @@ def training_initializer():
     method will automatically be called.  This function is designed to be used
     as a decorator on a factory.
 
+    Although the name suggests these initializers are training-related, they have
```
piercefreeman, line 13: apropo of my comment earlier, in this case let's think up a better name... Maybe "ml_initializers"

piercefreeman (approved):

andrewplumb (approved):

PR #116, opened 2021-01-19
Update `training_initializers` and `experiment_metrics`

salmanmashayekh:
* Call `training_initializers.init()` before any `bundle.fit()` (instead of just once at the beginning of the training).
* Remove `experiment_metrics` from the `training_initializers` and initialize them separately for training and evaluation

### DIFF
--- a/microcosm_sagemaker/commands/evaluate.py
+++ b/microcosm_sagemaker/commands/evaluate.py
@@ -32,6 +32,6 @@ def run_evaluate(
     input_data: InputData,
 ) -> None:
 
-    graph.training_initializers.init()
+    graph.experiment_metrics.init()
 
     graph.active_evaluation(graph.active_bundle, input_data)
--- a/microcosm_sagemaker/commands/train.py
+++ b/microcosm_sagemaker/commands/train.py
@@ -63,7 +63,7 @@ def run_train(
 ) -> None:
     root_output_artifact.init()
 
-    graph.training_initializers.init()
+    graph.experiment_metrics.init()
 
     root_output_artifact.save_config(graph.config)
 
--- a/microcosm_sagemaker/decorators.py
+++ b/microcosm_sagemaker/decorators.py
@@ -10,9 +10,6 @@ def training_initializer():
     method will automatically be called.  This function is designed to be used
     as a decorator on a factory.
 
-    Although the name suggests these initializers are training-related, they have
-    grown beyond that and they are called during both training and evaluation runs.
-
     """
     def decorator(func: Callable[[ObjectGraph], Any]):
         def factory(graph):
--- a/microcosm_sagemaker/metrics/wandb/store.py
+++ b/microcosm_sagemaker/metrics/wandb/store.py
@@ -3,7 +3,7 @@ import os
 from microcosm.api import binding, defaults
 from microcosm_logging.decorators import logger
 
-from microcosm_sagemaker.decorators import metrics_observer, training_initializer
+from microcosm_sagemaker.decorators import metrics_observer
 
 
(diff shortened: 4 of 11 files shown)


microcosm_sagemaker/commands/evaluate.py
```diff
@@ -32,6 +32,6 @@ def run_evaluate(
     input_data: InputData,
 ) -> None:
 
-    graph.training_initializers.init()
+    graph.experiment_metrics.init()
```
marcglobality, line 35: are there any implications of this change? e.g. do we want to fix the seeds when doing evaluation? (can't think of more, but maybe there are)

microcosm_sagemaker/commands/train.py
```diff
@@ -63,7 +63,7 @@ def run_train(
 ) -> None:
     root_output_artifact.init()
 
-    graph.training_initializers.init()
+    graph.experiment_metrics.init()
```
marcglobality, line 66: I guess this calls `fit_and_save_bundle` later on so shouldn't be a problem here?

microcosm_sagemaker/decorators.py
```diff
@@ -10,9 +10,6 @@ def training_initializer():
     method will automatically be called.  This function is designed to be used
     as a decorator on a factory.
 
-    Although the name suggests these initializers are training-related, they have
-    grown beyond that and they are called during both training and evaluation runs.
-
```
marcglobality, line 15: should we change the name then?

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -71,6 +70,9 @@ def _create_new_wandb_run(self, notes=None):
 
         # Injecting the wandb run path into the config
         self.graph.config.wandb.run_path = wandb_run.path
+        # Setting the run_path for the current instance.
+        # This prevents a new wandb instantiation if the `init` method is called again.
+        self.run_path = self.graph.config.wandb.run_path
```
marcglobality, line 75: not that it will change anything, but for me would be easier to read if assigning `wandb_run.path` as above (but totally personal style / minor)

marcglobality (approved):
LGTM. Added some small comments, but generally 👏

microcosm_sagemaker/decorators.py
salmanmashayekh: Why? They are still `training_initializers` and we may come up with other things to be registered in the future.

microcosm_sagemaker/commands/evaluate.py
salmanmashayekh: I cannot think of any use case where the seeds should be fixed for eval.

microcosm_sagemaker/decorators.py
marcglobality: well, given that you had a comment to justify why they are called like this, I thought was better to change the name. But it was an open question, I don't really have a strong opinion on this

microcosm_sagemaker/commands/evaluate.py
marcglobality: the seeds was just an example, but ok, if you cannot think of more, hopefully there aren't :D

pokey: If the evaluation is stochastic in some way, you wouldn't be able to reproduce it.  Not sure how likely this is, but doesn't seem like a big downside to seeding during eval.  Could factor out random seed code and call it from both `training_initializers` and `evaluation_initializers`, also calling metrics init during the latter.

But I imagine I lack some context these days 😄

PR #125, opened 2021-04-29
GLOB-52695: Make step param optional for wandb.

sstrasser-globality:
In experimental matcher, we are logging some charts at various time steps during evaluation. These changes allow us to not worry about setting the step parameter during logging.

### DIFF
--- a/microcosm_sagemaker/metrics/wandb/store.py
+++ b/microcosm_sagemaker/metrics/wandb/store.py
@@ -89,11 +89,8 @@ class WeightsAndBiases:
         return wandb_run
 
     def log_timeseries(self, **kwargs):
-        if "steps" in kwargs:
-            step = kwargs.pop("step")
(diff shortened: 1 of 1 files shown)


andrewplumb (approved):
approved contingent on passing tests

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -89,8 +89,11 @@ def _create_new_wandb_run(self, notes=None):
         return wandb_run
 
     def log_timeseries(self, **kwargs):
-        step = kwargs.pop("step")
-        self.wandb_run.log(kwargs, step=step)
+        if "steps" in kwargs:
```
salmanmashayekh, line 92: You can simplify it with the following:
```
step = kwargs.pop("step", None)
self.wandb_run.log(kwargs, step=step)
```

salmanmashayekh (approved):
LGTM

PR #130, opened 2021-05-12
GLOB-53866 Add a new loader to load command ling args into config

salmanmashayekh:
* Updates the `train` command to allow for extra unknown args --> It will eat up all of those args and pass it along the new loader
* Adds a new loader that parses command line args with this format: `--foo.bar=value`
* Also adds the test that goes with that.


microcosm_sagemaker/commands/train.py
```diff
@@ -4,19 +4,31 @@
...
 from microcosm_sagemaker.exceptions import raise_sagemaker_exception
 from microcosm_sagemaker.input_data import InputData
 
 
-@command()
+@command(
+    context_settings=dict(
+        ignore_unknown_options=True,
+    )
+)
```
pokey, line 31: This is obviously a bummer, but not sure I see a better approach given the constraints.  Does microcosm have a strict config mode that complains if you pass in a config entry not declared in defaults?  If so, you could catch bad arguments that way.  If not, then I think you just merge and try not to make any typos on the command line 😊

pokey (approved):
LGTM with a couple minor comments 😊

microcosm_sagemaker/tests/commands/train/test_command_line_config.py
```diff
@@ -0,0 +1,55 @@
...
+
+from hamcrest import assert_that, has_entries
+
+from microcosm_sagemaker.commands import train
+from microcosm_sagemaker.testing.cli_test_case import CliTestCase
+from microcosm_sagemaker.tests.fixtures import get_fixture_path
+from microcosm_sagemaker.tests.mocks import mock_app_hooks
+
+
+class TestTrainCommandLineConfig(CliTestCase):
```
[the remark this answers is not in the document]
salmanmashayekh: We barely ever use the config from file. Also, I assume that is something already tested in `merge`, right?

microcosm_sagemaker/commands/train.py
salmanmashayekh: Not sure about this, but this is certainly prone to typos. I will ask around to see if it is possible. 
A hacky way would be to create the `graph` with and without the command line config and make sure they have the same keys. But I am not sure it is worth doing @pokey .

piercefreeman: A safe mode's a good idea - practically speaking it might also be relatively easy to implement. Just make sure that all the items that are being processed by the `loaders` are located within this var already:

https://github.com/globality-corp/microcosm/blob/90291dadcbf535e7e245bfda15f3c663d9b176dd/microcosm/decorators.py#L34

In either case this should be a PR addition to microcosm instead of here, so going to approve this one for now.

piercefreeman (approved):

microcosm_sagemaker/commands/train.py
pokey: Agreed. Hope you guys are doing well 😊

PR #133, opened 2021-08-06
Make random.seed typed

tomaszpietruszka-globality:
Make random.seed typed so that it can be set from the command line


marcglobality (approved):

salmanmashayekh (approved):
LGTM

PR #135, opened 2021-08-18
Add pytest logging support

piercefreeman:
To make it easier to retrieve pytest results, introduce support for uploading to wandb.


microcosm_sagemaker/testing/pytest_sagemaker.py
```diff
@@ -0,0 +1,107 @@
...
+    # Get terminal contents and delete file
+    config._pytestsessionfile.seek(0)
+    sessionlog = config._pytestsessionfile.read()
+    config._pytestsessionfile.close()
+
+    del config._pytestsessionfile
+
+    # Undo our patching in the terminal reporter
+    tr = config.pluginmanager.getplugin('terminalreporter')
+    del tr._tw.__dict__['write']
```
pokey, line 55:
```suggestion
    terminal_reporter = config.pluginmanager.getplugin('terminalreporter')
    del terminal_reporter._tw.__dict__['write']
```

microcosm_sagemaker/testing/pytest_sagemaker.py
```diff
@@ -0,0 +1,107 @@
...
+        run_path = configuration.get("wandb", {}).get("run_path", {})
+
+    if not run_path:
+        warning("Unable to upload logging file to `wandb`: no runpath found")
+        return
+
+    # Load the wandb run that is linked to this artifact
+    api = wandb.Api()
+    run = api.run(run_path)
+    warning(f"Writing results to path: {run_path}")
```
pokey, line 101: Why is this a warning? Also, this isn't the path you're writing to right? Message might be slightly misleading but could be confused

microcosm_sagemaker/testing/pytest_sagemaker.py
```diff
@@ -0,0 +1,107 @@
...
+
+    with NamedTemporaryFile(prefix="", suffix=".log", mode="w+") as file:
+        file.write(contents)
+        file.seek(0)
```
pokey, line 105: Is this the same as

```suggestion
        file.flush()
```

pokey (approved):
This is looking great @piercefreeman. Just left a few minor comments 😊

README.md
```diff
@@ -61,3 +61,11 @@ class MyClassifier(Bundle):
...
 
 Note that the `step` keyword argument must be provided for logging time-series.
+
+## Artifact Tests
+
+If you want to report your artifact tests to wandb, add the following line to the top of your `conftest.py`:
+
+```
+pytest_plugins = 'pytest_sagemaker'
+```
```
marcglobality, line 71: this wasn't trivial for me, maybe add a reference to https://docs.pytest.org/en/6.2.x/plugins.html#requiring-loading-plugins-in-a-test-module-or-conftest-file?

microcosm_sagemaker/testing/pytest_sagemaker.py
marcglobality: maybe just changing "path" by "run" is enough here. I agree with the warning comment, but maybe that's the only way to get this shown ¿?

marcglobality (approved):
Just added some nits for better readability. Thanks for this!

microcosm_sagemaker/testing/pytest_sagemaker.py
piercefreeman: Flush will just write whatever current contents there are to the file (which should have already been done). `seek` resets the file pointer to the open file, which we need to stream from the beginning when we call the `upload_file` handler.

pokey: But won't `upload_file` just reopen the file?  You're just passing in its name, so I'm not sure how it would reuse the open file handle

piercefreeman: @pokey It is unintuitive, but I think it has to do with underlying file pointers. Code to reproduce:

```
def call_path(path):
  with open(path, "r") as file:
    print("AWS sees", file.read())

with NamedTemporaryFile(prefix="", suffix=".log", mode="w+") as file:
  file.write("Testing line")
  call_path(file.name)

> ('AWS sees', '')
```

PR #140, opened 2022-01-31
Update requirements to remove obsolete pinning

uri-miron-globality:
Remove pinning for python-dateutil, which forces old version of botocore… The specific conflict it raised in my project was with dvc[s3].


setup.py
```diff
@@ -25,8 +25,7 @@
         "click>=7.0",
         "microcosm>=2.12.0",
         "microcosm-flask[metrics]>=2.8.0",
-        # See: https://github.com/boto/botocore/pull/1910
-        "python-dateutil<3.0.0",
+        "python-dateutil",
```
marcglobality, line 28: they still have this in their reqs. In other words, the change on that PR it's still present in master. Are you sure it's not needed? if so, could you create an issue in botocore to ask them to remove?

marcglobality (approved):
Please create the issue in botocore for the benefit of the community :)

setup.py
uri-miron-globality: Yes, looks like a meaningless change. Will close the PR.

PR #141, opened 2022-02-09
Make LSTM/GRU deterministic

marcglobality:
As talked in  Provider data OKR sync:

* make LSTM/GRU deterministic
* update README


tomaszpietruszka-globality (approved):
🎉

PR #147, opened 2022-10-12
runserver threading flag

salmanmashayekh:
`globality-taxonomies` doesn't have support for local multi-threaded flask serving, since you can't re-use taxonomy entities on other threads once spawned on the master. Here we add a CLI flag to `runserver` to control whether it launches flask in single threaded or multi-threaded mode. It defaults to true to maintain existing behavior. 

It is a duplicate of an old PR that was not merged (unknown reason): https://github.com/globality-corp/microcosm-sagemaker/pull/98

Also included in this PR:
* Updating build files
* Not using `git-flow` to release the library anymore 
* Removing old classes using old test conventions (everything being used before the `pipeline_harness`)
* Upgrading to python 3.9


tomaszpietruszka-globality (approved):
LGTM!

marcglobality (approved):

marcglobality:
> `globality-taxonomies` doesn't have support for local multi-threaded flask serving, since you can't re-use taxonomy entities on other threads once spawned on the master. Here we add a CLI flag to `runserver` to control whether it launches flask in single threaded or multi-threaded mode. It defaults to true to maintain existing behavior.
> 
> I tried it locally and it works fine.
> 
> It is a duplicate of an old PR that was not merged (unknown reason): #98

I cannot claim I fully understand this, but if you `can't re-use taxonomy entities on other threads once spawned on the master`, should not the default be False so there are no threads and we don't run into issues? That being said, the default is False [here](https://werkzeug.palletsprojects.com/en/2.2.x/serving/#werkzeug.serving.run_simple), so yeah defo I don't understand it :D

marcglobality:
we could also change the target of this to `master` and apply these [changes](https://github.com/globality-corp/crea-commons/pull/66/files#diff-a057a3936e798488200642c4048bbdcfde657778e2dc6a9e1bcc66f242ec10b6R16-R24), so removing git-flow forever and ever

tomaszpietruszka-globality:
👍 for the `default=False`

README.md
```diff
@@ -72,7 +72,7 @@ pytest_plugins = 'pytest_sagemaker'
 ```
```
marcglobality, line 72: maybe we can use this chance to update the README and explain the two apps that come with it? `train` `serve` and `evaluate` ?

entrypoint.sh
```diff
@@ -24,22 +24,22 @@
...
 if [ "$1" = "test" ]; then
-   apt-get update
-   apt-get install -y --no-install-recommends build-essential
+   pip install --upgrade pip
    # Install standard test dependencies; YMMV
    pip --quiet install \
-       .[test] nose PyHamcrest coverage
-   exec nosetests
+       .[test]
+   nosetests
```
marcglobality, line 31: can we move to pytest as well? should be just adding to build.json and regenerating, as in https://github.com/globality-corp/web-annotation-extractor/blob/master/.globality/build.json#L67

setup.py
```diff
@@ -84,6 +84,12 @@
             "parameterized>=0.7.0",
             "torch>=1.1.0",
             "wandb>=0.10",
+            "nose>=1.3.6",
```
marcglobality, line 87: if we move to pytest, we can remove this

marcglobality (commented):
Added two major buckets of comments:
- I'm unsure the removed tests are not needed
- Move to pytest, which seems very quick to do

Approving anyways as both can be done at another PR

.circleci/config.yml
```diff
@@ -240,17 +249,19 @@ jobs:
...
+            echo "    jfrog " >> ~/.pypirc
             echo >> ~/.pypirc
-            echo "[pypi]" >> ~/.pypirc
-            echo "repository = https://upload.pypi.org/legacy/" >> ~/.pypirc
-            echo "username = __token__" >> ~/.pypirc
-            echo "password = ${PYPI_AUTH_TOKEN}" >> ~/.pypirc
+            echo "[jfrog]" >> ~/.pypirc
+            echo "repository = https://globality.jfrog.io/globality/api/pypi/pypi" >> ~/.pypirc
+            echo "username   = $JFROG_USERNAME" >> ~/.pypirc
+            echo "password   = $JFROG_PASSWORD" >> ~/.pypirc
```
tomaszpietruszka-globality, line 257: I think we should still publish this into pypi, no?

.globality/build.json
```diff
@@ -24,9 +24,19 @@
     "name": "microcosm-sagemaker",
     "pypi": {
       "enabled": true,
-      "repository": "pypi"
```
tomaszpietruszka-globality, line 27: perhaps worth keeping?

tomaszpietruszka-globality (approved):
LGTM again/still

PR #148, opened 2022-10-14
Merge develop into master

tomaszpietruszka-globality:
We are done with git-flow. 

This is to sync master with develop ahead of the other PR, which will adjust the settings


tomaszpietruszka-globality:
We should remember not to squash when merging this

PR #152, opened 2022-11-15
Glob 69972 redesign model artifact tests

jvonalbade-globality:
The artifact performance test, which was previously a pytest in e.g. qna-choice-suggestion, has been moved into the train script in microcosm-sagemaker. This allows for testing model performance on the same hardware as training rather than on the same hardware as deployment. See https://globality.atlassian.net/wiki/spaces/AI/pages/2859597825/Re-Designing+Model+Artifact+Tests for further detail.

This test assures that the model meets minimal requirements for e.g. precision and recall. After training, the model is evaluated on test data and a number of assertions are run. If the assertions are not met, an error is thrown, but the model is still saved for debugging purposes. As in the Pytest, if the test data is "skipped," the test is not run. There is a new command line argument, set to True by default, that allows one to skip this test.

This updated version of microcosm-sagemaker will be used first for a revised qna-choice-suggestion, then later for a revised service-offering-classifier.

When performance meets the assertions on the test data, we see training executes as expected:

Screen Shot 2022-11-15 at 3 25 53 PM

And when we change the assertions so that performance does not meet them (here by increasing the minimum recall), we get an error as expected:

Screen Shot 2022-11-15 at 3 28 21 PM

And finally if we pass the command line argument to bypass the test, training completes without performing the assertions:

Screen Shot 2022-11-15 at 3 38 26 PM

See https://github.com/globality-corp/qna-choice-suggestion/pull/63 for the corresponding pull request in qna-choice-suggestion.


microcosm_sagemaker/commands/train.py
```diff
@@ -69,6 +74,15 @@ def main(configuration, input_data, output_artifact, auto_evaluate, testing, com
         if auto_evaluate:
             run_evaluate(graph, input_data)
 
+        if metric_assertions:
+            graph = create_evaluate_app(
```
salmanmashayekh, line 78: Can we add a comment here why we should recreated the graph, as opposed to using the existing graph?

salmanmashayekh (approved):
LGTM!
If we don't already have a test that does "metric-assertions", can we add one?

marcglobality:
could you link from this PR to the one in qna? so we have both cross-linked

microcosm_sagemaker/factories.py
```diff
@@ -20,6 +20,12 @@ def configure_active_evaluation(graph):
     return getattr(graph, graph.config.active_evaluation)
 
 
+def configure_test_evaluation(graph):
+    if not getattr(graph.config, "test_evaluation", ""):
+        return None
+    return getattr(graph, graph.config.test_evaluation)
```
marcglobality, line 26: appreciate you just copied the above pattern, but what's exactly is done here? maybe @salmanmashayekh is better to answer

microcosm_sagemaker/commands/evaluate.py
```diff
@@ -41,3 +41,15 @@ def run_evaluate(
     graph.experiment_metrics.init()
 
     graph.active_evaluation(graph.active_bundle, input_data)
+
+
+def run_evaluate_on_test(
```
marcglobality, line 46: I find very confusing to have an `active_evaluation` and a `test_evaluation`. What's the purpose of each?

microcosm_sagemaker/tests/commands/train/test_train_distributed_cluster.py
```diff
@@ -52,6 +53,7 @@ def test_train_main_process(self) -> None:
                     "--output-artifact",
                     output_artifact_path,
                     "--testing",
+                    "--no-metric-assertions"
```
marcglobality, line 56: maybe tests should cover with assertions and without?

marcglobality (commented):
Added some comments. Not approving yet as I want to clarify the diff btw `active_evaluation` and `test_evaluation`, both for my own and in comments/docs

microcosm_sagemaker/commands/train.py
jvonalbade-globality: Done - just added in your comment from the writeup, Salman. @marcglobality Basically we do not have to recreate the graph, but by doing do we can also test saving/loading.

microcosm_sagemaker/commands/train.py
```diff
@@ -69,6 +74,15 @@ def main(configuration, input_data, output_artifact, auto_evaluate, testing, com
...
+                extra_loader=load_from_dict(
+                    extra_config,
+                    root_input_artifact_path=output_artifact.path),
```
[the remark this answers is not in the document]
salmanmashayekh: We do have the graph from training app, but we intentionally recreate it to make sure "loading" works as expected. I knew the answer, and as @jvonalbade-globality did, I wanted a comment that explains it.

microcosm_sagemaker/commands/evaluate.py
salmanmashayekh: Let me answer this question and the one above @marcglobality. As you know, we have `active_bundle` and `active_evaluation`. The former one defines the master bundle, and the latter defines the evaluation bundle for **train and dev** sets. 

In this update, we also want to define (in each AI service) what the evaluation bundle for the **test** set is. That is why we have introduced `test_evaluation`. Ideally, we would name them `dev_evaluation` (instead of `active_evaluation`) and `test_evaluation`, but did not want to change the existing name. 

What `configure_test_evaluation` does is exactly like configuring active evaluation (for train/dev) and active bundle. It uses the config to know what eval bundle is used for the test set and attaches it to the graph as `graph.test_evaluation`.

microcosm_sagemaker/tests/commands/train/test_train_distributed_cluster.py
salmanmashayekh: I think we already have tests without metric assertions.

jvonalbade-globality: I changed the second test in this file to run with assertions and noted in a comment that we need the "--no-metric-assertions" tag on the first or it will fail, since there is no artifact produced.

microcosm_sagemaker/commands/evaluate.py
marcglobality: I see. In the WAE we don't really evaluate both the develop and test datasets, as it takes a long time. But appreciate that's how should be done and it's the correct approach

marcglobality (approved):

tomaszpietruszka-globality (approved):
LGTM, with the 1 optional comment.

And 👍 for Marc's comment about using the formatter

microcosm_sagemaker/factories.py
```diff
@@ -20,6 +20,12 @@ def configure_active_evaluation(graph):
     return getattr(graph, graph.config.active_evaluation)
 
 
+def configure_test_evaluation(graph):
+    if not getattr(graph.config, "test_evaluation", ""):
```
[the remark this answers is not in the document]
tomaszpietruszka-globality: I see that we do. Please disregard

PR #156, opened 2023-01-10
GLOB-71669 spawn to fit

uri-miron-globality:
Add support for training a bundle from a separately-spawned process. 

To mark a bundle for training on a separate process, its `spawn_to_fit` property should be overriden to return True.

Closes [GLOB-71669]

[GLOB-71669]: https://globality.atlassian.net/browse/GLOB-71669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


microcosm_sagemaker/commands/train_one_bundle.py
```diff
@@ -0,0 +1,148 @@
...
+from microcosm_sagemaker.distributed_cluster import is_worker_process
+from microcosm_sagemaker.input_data import InputData
+from microcosm_sagemaker.metrics.wandb.wandb_config import WeightsAndBiasesConfig
+
+
+def find_binding_name(graph, obj) -> Optional[str]:
+    for name, obj_ in graph.items():
+        if obj_ is obj:
+            return name
+    return None
```
marcglobality, line 24: can't you reuse this one? https://github.com/globality-corp/microcosm-sagemaker/blob/e2c7dcc6586a6c88f9c0ac85270e61063cfa70c3/microcosm_sagemaker/bundle_traversal.py#L113

microcosm_sagemaker/commands/train_one_bundle.py
```diff
@@ -0,0 +1,148 @@
...
+    bundle.load(BundleInputArtifact(artifact_path / bundle_name))
+
+
+def load_bundle_dependencies(bundle, graph, artifact_path: Path):
+    orchestrator = SingleThreadedBundleOrchestrator(graph)
+    orchestrator(
+        bundle=bundle,
+        bundle_handler=partial(load_bundle, graph=graph, artifact_path=artifact_path),
+        dependencies_only=True,
+    )
```
marcglobality, line 42: it seems like you could reuse this one https://github.com/globality-corp/microcosm-sagemaker/blob/e2c7dcc6586a6c88f9c0ac85270e61063cfa70c3/microcosm_sagemaker/bundle_traversal.py#L17. If the only thing you want different is to load deps but not the bundle itself, just modify that one with some flags but I think it's better to reuse components when we can

microcosm_sagemaker/constants.py
```diff
@@ -25,3 +25,6 @@ class SagemakerPath:
 # NOTE: We can only use up to 10 dimensions:
 # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudwatch.html
 CLOUDWATCH_MAX_DIMENSIONS = 10
+
+
+WANDB_ENV_VARIABLES = ["WANDB_ENTITY", "WANDB_RUN_GROUP", "WANDB_NAME"]
```
marcglobality, line 30: what is each of these for? will you write some docs in the README and/or Confluence?

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,108 @@
...
+
     def __init__(self, graph):
         self.graph = graph
         self.enable = (
             # If it is not explicitly enabled or disabled, enable if not testing.
             graph.config.weights_and_biases.enable or not graph.metadata.testing
         )
+        if self.enable:
+            try:
+                wandb.run
```
marcglobality, line 37: this line allows you to say that wandb is not installed? shouldn't be when importing wandb?

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,108 @@
...
         self.wandb_run = None
 
-    def init(self):
-        # Sometimes, the entity in the wandb run path appears as None.
-        # To make sure this does not happen, we explicitly set the wandb entity.
-        # https://docs.wandb.com/library/environment-variables
-        os.environ["WANDB_ENTITY"] = "globality"
+        # settings used by init
+        self.run_config: Optional[WeightsAndBiasesConfig] = None
+        self.from_server: bool = False
```
marcglobality, line 48: I don't understand this one. Maybe a comment would be useful here

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,108 @@
...
+            run_config (Optional[WeightsAndBiasesConfig]): wandb run configuration
+            from_server (bool, optional): Whether to create a new run. Set to true to push data
+                into the run after training (e.g. for artifact tests). Defaults to False.
+        """
         # Only initialize wandb if it is enabled
         if not self.enable:
             return
 
+        if self.wandb_run is not wandb.run:
+            self.wandb_run = wandb.run
```
marcglobality, line 82: why not just 82 and remove the if?

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,108 @@
...
-            self.wandb_run = wandb.Api().run(path=self.run_path)
-            self.logger.info(f"The existing Weights & Biases run was loaded: {self.run_path}")
-        except CommError:
-            # If the previous run cannot be found:
-            #   - Start a new run instead
-            #   - Add a note to the new run that this run is related to a previous run
-            self.logger.warning(f"Could not find run: {self.run_path}")
-            self.wandb_run = self._create_new_wandb_run(
-                notes=f"Related to previous run: {self.run_path}",
+        if not self.from_server:
```
marcglobality, line 89: can you add comments to the blocks to explain what they are doing?

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,108 @@
...
+        self.logger.info(f"wandb server config: run_path = {run_path}")
+
+    def init(self) -> None:
+        """ Initialize wandb run.
+
+        Args:
+            run_config (Optional[WeightsAndBiasesConfig]): wandb run configuration
+            from_server (bool, optional): Whether to create a new run. Set to true to push data
+                into the run after training (e.g. for artifact tests). Defaults to False.
+        """
```
marcglobality, line 76: (nit) these are not really args from here. They could be better in the init, I think

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,108 @@
...
+            run_path = f"{self.run_config.entity}/{self.project_name}/{self.run_config.id}"
+            try:
+                self.wandb_run = wandb.Api().run(path=run_path)
+                self.logger.info(f"The existing Weights & Biases run was loaded: {run_path}")
+            except CommError:
+                # If the previous run cannot be found:
+                #   - Start a new run instead
+                #   - Add a note to the new run that this run is related to a previous run
+                self.logger.warning(f"Could not find run: {run_path}")
+                new_id = f"{self.run_config.id}_{''.join(random.choices(string.ascii_lowercase + string.digits, k=4))}"
```
marcglobality, line 120:
```suggestion
                new_id = f"{self.run_config.id}_{str(uuid.uuid4())[:4]}"
```
?

microcosm_sagemaker/metrics/wandb/wandb_config.py
```diff
@@ -0,0 +1,37 @@
...
+BASE_NAME_FOR_WANDB = "BASE_NAME_FOR_WANDB"
+TRAINING_JOB_NAME = "TRAINING_JOB_NAME"  # defined by sagemaker
+
+
+@dataclass
+class WeightsAndBiasesConfig:
+    entity: Optional[str] = None
+    group: Optional[str] = None
+    name: Optional[str] = None
+    id: str = field(default_factory=lambda: "".join(random.choices(string.ascii_lowercase + string.digits, k=8)))
```
marcglobality, line 19:
```suggestion
    id: str = field(default_factory=lambda: str(uuid.uuid4())[:8]
```
?

microcosm_sagemaker/tests/metrics/test_wandb_run_path.py
```diff
@@ -23,11 +23,15 @@
 
 @dataclass
 class MockedWandb:
-    """
-    Mocks the wandb run object with a path attribute.
```
marcglobality, line 27: keep "Mocks the wandb run object" ?

marcglobality (commented):
Added some comments. Two high level comments:
1. Adding info the README and link to a Confluence page would help. Also maybe link the SOC PR where you leverage this
2. Add tests for a bundle trained in a subprocess?

uri-miron-globality:
> Added some comments. Two high level comments:
> 
> 1. Adding info the README and link to a Confluence page would help. Also maybe link the SOC PR where you leverage this
> 2. Add tests for a bundle trained in a subprocess?

@marcglobality, regarding your first comment, will add explanation to README. Regarding the second, I don't see a way to do this that is worth the extra code complexity. The problem is we would need to extend the mocking logic [here](https://github.com/globality-corp/microcosm-sagemaker/blob/e2c7dcc6586a6c88f9c0ac85270e61063cfa70c3/microcosm_sagemaker/tests/mocks.py#L11) to work on a new process.

microcosm_sagemaker/commands/train_one_bundle.py
```diff
@@ -0,0 +1,148 @@
...
+
+def train_one_bundle(
+    bundle_name: str,
+    input_data_path: Path,
+    artifact_path: Path,
+    testing: bool,
+    wandb_entity: Optional[str],
+    wandb_group: Optional[str],
+):
+    logger = getLogger("train_one_bundle")
```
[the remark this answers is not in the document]
uri-miron-globality: Actually, I think the logger creation should be a few lines later, after the graph is created, to ensure it uses the same logging config.

microcosm_sagemaker/bundle_traversal.py
```diff
@@ -37,6 +41,47 @@ def load(bundle):
...
 
 
+def _spawn_fit_and_save_bundle(
+    bundle: Bundle,
+    bundle_name: str,
+    input_data_path: Path,
+    artifact_path: Path,
+    testing: bool,
+    graph: ObjectGraph,
+) -> None:
```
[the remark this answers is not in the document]
uri-miron-globality: There aren't really sensible defaults for the arguments, but made the other suggested changes

microcosm_sagemaker/commands/train_one_bundle.py
uri-miron-globality: 👍

microcosm_sagemaker/commands/train_one_bundle.py
uri-miron-globality: 👍

microcosm_sagemaker/constants.py
uri-miron-globality: These aren't needed, removed them

microcosm_sagemaker/metrics/wandb/store.py
uri-miron-globality: 👍

microcosm_sagemaker/metrics/wandb/store.py
uri-miron-globality: Not sure why that would be better.. we don't need a UUID here, just some random string

microcosm_sagemaker/tests/metrics/test_wandb_run_path.py
uri-miron-globality: 👍

microcosm_sagemaker/metrics/wandb/store.py
marcglobality: just for code brevity

microcosm_sagemaker/bundle.py
```diff
@@ -12,6 +12,11 @@ class Bundle(ABC):
     """
     predict: Callable
 
+    @property
+    def spawn_to_fit(self) -> bool:
+        """ Flag indicating whether fit should be run on a separate process. """
```
marcglobality, line 17: I wonder if worth to explain the default here

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,106 @@
...
-        # To make sure this does not happen, we explicitly set the wandb entity.
-        # https://docs.wandb.com/library/environment-variables
-        os.environ["WANDB_ENTITY"] = "globality"
+        # Settings used by init to configure the wandb run. These need to be set before the call to
+        # init() if non-default values are needed. A true value in `from_server` is used for joining an
+        # existing (completed) run, typically from a different machine.
+        self.run_config: Optional[WeightsAndBiasesConfig] = None
+        self.from_server: bool = False
+
+    def set_from_server_config(self):
```
marcglobality, line 52: docstring?

microcosm_sagemaker/metrics/wandb/wandb_config.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def is_main(self) -> bool:
+        assert self.name is not None
+        return self.name.endswith("__main")
+
+    @classmethod
+    def create_main_config(cls, create_group: bool, entity: Optional[str] = "globality") -> "WeightsAndBiasesConfig":
```
marcglobality, line 26: docstring to explain main_config vs the other configs?

marcglobality (approved):
Just added some more minor comments. LGTM

microcosm_sagemaker/metrics/wandb/store.py
uri-miron-globality: @marcglobality, wandb is an optional install for microcosm-sagemaker, but there is nothing preventing you from doing `graph.use("wandb")` even if you didn't install it.

marcglobality: ok, I see that I didn't understand what was going on. Still, I find it confusing the way it's written. Why do you need to do `wandb.run` if you just want to check if `wandb` exists? + the exception could be `NameError` ?

marcglobality: my point was not about the characters but about being explicit to the reader :)

microcosm_sagemaker/metrics/wandb/wandb_config.py
tomaszpietruszka-globality: Why do we have to generate the ID ourselves? I think wandb does it by default?

microcosm_sagemaker/metrics/wandb/wandb_config.py
```diff
@@ -0,0 +1,37 @@
...
+
+BASE_NAME_FOR_WANDB = "BASE_NAME_FOR_WANDB"
+TRAINING_JOB_NAME = "TRAINING_JOB_NAME"  # defined by sagemaker
+
+
+@dataclass
+class WeightsAndBiasesConfig:
+    entity: Optional[str] = None
+    group: Optional[str] = None
+    name: Optional[str] = None
```
tomaszpietruszka-globality, line 18: Wouldn't it make sense to include the project name here as well?

tomaszpietruszka-globality (approved):
LGTM, with a few comments - and I'll second Marc's requests for docstrings.

microcosm_sagemaker/metrics/wandb/store.py
uri-miron-globality: Regarding the server terminology, it's meant to emphasize that this is not resuming a local run. Also, the original logic relied on the config file state, which is hard to reason about in a multiple process setting. I think it's better for the application to specify explicitly when this connection mode is needed.

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,106 @@
...
         self.bundle_and_dependencies_config_extractor = self.graph.bundle_and_dependencies_config_extractor
         self.active_bundle = getattr(graph, graph.config.active_bundle)
-        self.run_path = getattr(graph.config.wandb, "run_path", None)
         self.wandb_run = None
 
-    def init(self):
-        # Sometimes, the entity in the wandb run path appears as None.
-        # To make sure this does not happen, we explicitly set the wandb entity.
-        # https://docs.wandb.com/library/environment-variables
-        os.environ["WANDB_ENTITY"] = "globality"
```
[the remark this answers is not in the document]
uri-miron-globality: It's being set explicitly in `wandb.init` (which has higher priority than env variables).

microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -20,80 +24,106 @@
...
+        # Settings used by init to configure the wandb run. These need to be set before the call to
+        # init() if non-default values are needed. A true value in `from_server` is used for joining an
+        # existing (completed) run, typically from a different machine.
+        self.run_config: Optional[WeightsAndBiasesConfig] = None
```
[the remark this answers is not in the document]
uri-miron-globality: It can't easily be in `__init__` since it's graph bound. Adding it to `init` breaks the `ExperimentMetrics` interface which assumes there could be other registered observers.

microcosm_sagemaker/metrics/wandb/wandb_config.py
uri-miron-globality: The wandb integration with sagemaker overrides this ID through the `WANDB_RUN_ID` env variable, which causes different behavior in sagemaker than a normal EC2 environment (found it later that newer versions of wandb fix this with name mangling, but I would rather leave this explicit)

microcosm_sagemaker/metrics/wandb/store.py
marcglobality: what about `load_existing_remote_run` ? still, I miss a docstring here

microcosm_sagemaker/metrics/wandb/wandb_config.py
marcglobality: if the latest versions of wandb work, why not set a lowerbound and leave support to them

microcosm_sagemaker/metrics/wandb/store.py
uri-miron-globality: Added explanation -- it's not displaying in github

README.md
```diff
@@ -86,3 +86,12 @@ To support "distributed" training - with multiple processes (like pytorch `Distr
...
 saving artifacts.
+
+## Training Bundles on Separate Processes
+
+In some circumstances, it is beneficial for the training of a bundle to take place on a separate 
+process. An example of this is when using DDP (or a DDP-related mechanism such as DeepSpeed) for multi-GPU 
+training. This training mode may be enbled by setting the `spawn_to_fit` property of the `Bundle` to true
+(note that the main process with block while the spawned process runs `fit` -- the purpose is to allow for 
+more efficient parallelism in the training of each individual bundle, not for the concurrent training of 
+multiple bundles).
```
marcglobality, line 97:
```suggestion
In some circumstances, it is beneficial for the training of a bundle to take place on a separate 
process. An example of this is when using DDP (or a DDP-related mechanism such as DeepSpeed) for multi-GPU 
training. This training mode may be enabled by setting the `spawn_to_fit` property of the `Bundle` to true
(default is to False). Note that the main process will block while the spawned process runs `fit` -- the purpose is to allow for more efficient parallelism in the training of each individual bundle, not for the concurrent training of 
multiple bundles.
```

marcglobality (commented):
A few more comments

microcosm_sagemaker/metrics/wandb/store.py
uri-miron-globality: 👍

microcosm_sagemaker/metrics/wandb/wandb_config.py
uri-miron-globality: 👍

setup.py
```diff
@@ -28,6 +28,8 @@
         # See: https://github.com/boto/botocore/pull/1910
         "python-dateutil<3.0.0",
         "joblib>=0.15",
+        "ipykernel",
```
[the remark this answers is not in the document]
uri-miron-globality: Not needed, will remove

microcosm_sagemaker/commands/train_one_bundle.py
```diff
@@ -0,0 +1,117 @@
...
+    prelude = f"{bundle_name} spawned process training"
+    configuration_path = artifact_path / ARTIFACT_CONFIGURATION_PATH
+    if not configuration_path.exists():
+        raise ValueError(f"{prelude}: Could not find config file {configuration_path}!")
+    with open(configuration_path, "r") as fp:
+        extra_loader = load_from_dict(json.load(fp))
+    graph = create_train_app(
+        extra_loader=extra_loader,
+        testing=testing,
+    )
```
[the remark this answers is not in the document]
uri-miron-globality: Let's leave it for a future PR

microcosm_sagemaker/commands/train_one_bundle.py
```diff
@@ -0,0 +1,117 @@
...
+
+def train_one_bundle(
+    bundle_name: str,
+    input_data_path: Path,
+    artifact_path: Path,
+    testing: bool,
+    wandb_entity: Optional[str],
+    wandb_group: Optional[str],
+):
+    prelude = f"{bundle_name} spawned process training"
```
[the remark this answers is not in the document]
uri-miron-globality: 👍

PR #157, opened 2023-02-07
add wandb project name setting

uri-miron-globality:
Add the ability to set the wandb project name.


marcglobality (commented):
Could you link a PR (from SOC I suppose) where this is used and working?

uri-miron-globality:
> Could you link a PR (from SOC I suppose) where this is used and working?

@marcglobality, I don't have a PR that uses it, it's mainly to for experiments (to not use the same `wandb` project as production).

marcglobality:
As discussed offline, here's an example on how to use it:
```
from fancy_classifier.app_hooks.train.app import create_app
from microcosm.loaders import load_from_dict

CONFIG = dict(
    weights_and_biases=dict(project_name="fancy-classifier", enable=True),
)

graph = create_app(
    debug=True,
    testing=False,
    extra_loader=load_from_dict(CONFIG),
)
```

marcglobality (approved):

salmanmashayekh (approved):
LGTM

PR #159, opened 2023-03-02
Glob 72600/migrate to drone

marcglobality:


setup.py
```diff
@@ -3,7 +3,7 @@
 
 
 project = "microcosm-sagemaker"
-version = "0.9.5"
+version = "0.1.0"
```
salmanmashayekh, line 6: Do you mean `0.10.0`?

salmanmashayekh (approved):
LGTM with a comment about the version number

jvonalbade-globality (approved):
LGTM

setup.py
marcglobality: I had to do this because of this https://github.com/globality-corp/globality-build/blob/1a09cba810fc96de427a0e5398141fb7c215748f/globality_build/templates/drone/python-library/.drone.yml#L87

PR #161, opened 2023-05-01
New training pipeline improvements

salmanmashayekh:
Makes small fixes to the new training pipeline:
* The logic to determine whether a bundle needs to train was not working properly when bundles were trained on spawn processes. That is fixed in this PR. 
* The logic to create artifacts dir from wandb cloud was not checking if the new pipeline is enabled in the yaml file. That is fixed in this PR.


microcosm_sagemaker/metrics/wandb/store.py
```diff
@@ -90,6 +90,14 @@ def assert_only_one_new_artifact_id(self):
             f"but this one had {new_artifact_ids}"
         )
 
+        # We run this to trigger the assertion error within the find_run_for_artifact function.
+        # It fails if the bundle is tagged with an artifact id that exists, but the bundle artifact
+        # is not found.
+        _ = [
+            find_run_for_artifact(self.project_name, artifact_id, bundle_name)
+            for bundle_name, artifact_id in self.artifacts_config["bundles"].items()
+        ]
```
pokey, line 99: Why not just use a regular `for` loop?

pokey (commented):
👋😊

microcosm_sagemaker/metrics/wandb/store.py
salmanmashayekh: Hey @pokey ! Nice to hear from you! :)
Yeah, a regular `for` would have been cleaner. Got rid of this whole thing in new commits tho!

microcosm_sagemaker/bundle_traversal.py
```diff
@@ -108,11 +109,14 @@ def fit_and_save_bundle(
...
         artifact_id = artifacts_data["bundles"][bundle_name]
+        # We intentionally do not pass the bundle_name to find_run_for_artifact since we don't want the
+        # assertion for bundle artifact being in the run to be triggered.
         run_id = find_run_for_artifact(project, artifact_id)
-        if run_id is not None:
+        if not bundle_needs_to_fit(run_id, project, bundle_name):
             # bundle does not need to be trained - instead, we copy its artifacts from the cloud to local and load it
             logger.info(f"The artifact for bundle `{bundle_name}` found, training will be skipped...")
             bundle_dir = root_output_artifact.path / bundle_name
+            assert isinstance(run_id, str)
```
marcglobality, line 119: maybe this could go inside `copy_run_artifact_to_local` ?

microcosm_sagemaker/bundle_traversal.py
```diff
@@ -108,11 +109,14 @@ def fit_and_save_bundle(
     if artifacts_enabled:
         # check if the bundle needs to be trained
         artifact_id = artifacts_data["bundles"][bundle_name]
+        # We intentionally do not pass the bundle_name to find_run_for_artifact since we don't want the
+        # assertion for bundle artifact being in the run to be triggered.
```
marcglobality, line 113: this comment is a bit confusing for someone not extremely familiar to the code. Where is the assertion for bundle artifact? at least mention the method name

Also, I think it'd help something like (maybe I misunderstood, so nvm if that's the case)
```
This is especially required when the bundle is trained in a new process (via the `spawn_to_fit` feature).
In such a case, we want the assertion in the run of the main process, not the run of the spawned one.
```

marcglobality (approved):
LGTM

microcosm_sagemaker/bundle_traversal.py
salmanmashayekh: The comment is not relevant anymore, removed.

